Issue #30, opened 2014-12-15
Dns.Packet.parse_rdata incorrectly parses SRV


infidel:
The code is:

```ocaml
    | RR_SRV ->
        BE.(
          SRV (get_uint16 buf 0, (* prio *)
               get_uint16 buf 2, (* weight *)
               get_uint16 buf 4, (* port *)
               parse_name names (base+6) buf |> stop
          ))
```

But the "base" argument to parse_name is ignored for parsing (it only affects the return value), so it doesn't work. The most obvious fix is to shift the buffer by 6 before passing to parse_name, and that would match most of the other RR types. I can submit a PR to that effect, including a unit test that triggers the bug (preferably after my other PR has been merged).

However, the values passed for "base" are inconsistent and the argument seems to be fairly pointless?

dsheets:
I think this can be closed now as 0.12.0 indicates a fix shipped. Please re-open if this is incorrect.

If fixing something like `base` is the reason this is still open, please open a new issue for just that.

PR #31, opened 2014-12-21
Fix bug in parsing of PTR, and add a test

lcdunstan:
Fixes PTR (#30) and AFSDB.

I've also changed the other calls to parse_name from parse_rdata to pass zero for "base", because passing anything other than zero is a bit misleading.

I still think that the "base" argument could probably be removed entirely to simplify the code, and parse_name could still return the length if there is some caller that uses it.


mor1:
yup :)


On 22 December 2014 at 11:37, Anil Madhavapeddy <notifications@github.com>
wrote:

> Aha! Yes, label compression, that old friend. Thanks...
>
> > On 21 Dec 2014, at 19:29, Luke Dunstan <notifications@github.com>
> wrote:
> >
> > Very sorry, but another unit test revealed that my code is wrong. The
> base parameter is also needed to calculate the absolute offset of every
> label for storage in the decompression hash table. I will fix it soon.
> >
> > --
> > Reply to this email directly or view it on GitHub <
> https://github.com/mirage/ocaml-dns/pull/31#issuecomment-67798796>.
> >
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/mirage/ocaml-dns/pull/31#issuecomment-67828089>.
>



-- 
Richard Mortier
mort@cantab.net

Issue #28, opened 2014-11-23
Support for mDNS?


infidel:
I am interested in implementing a basic Multicast DNS resolver/responder (RFC 6762), mainly as a learning exercise (and only in my spare time). I had not heard of mDNS until reading the Cubieboard2 tutorials on openmirage.org, but I've been reading the RFCs since then. The top-level algorithms would be best implemented in a separate "mdns" package rather than in "dns", but most of the packets and other data structures will be shared with "dns", with minor differences.

Are you open to receiving patches to allow "dns" to support mDNS and perhaps DNS-SD?

As an example, DNS has a 16-bit resource record class field, while mDNS uses a 15-bit RR class plus a 1-bit "cache flush" bit (with the exception of some types like OPT). To start with, this requires a change to Dns.Packet.parse_rr to mask out bit 15 before calling int_to_rr_class, and to return the cache flush bit in the result.

infidel:
OK taking your hint, I managed to get Jitsu going with normal DNS. It can resume/suspend a unikernel (I don't know how to get "shutdown" to work with a unikernel though). I understand that Jitsu is acting as a forwarding DNS resolver. Is it something that you would expect to use on the internet directly or is it more for use on a private LAN/WAN?

As for mDNS, which is for LANs, the concept of a "forwarder" doesn't exist but Jitsu could listen for queries of A records much like DNS. In the case where the unikernel started by Jitsu also implements a normal mDNS responder then it may be better if Jitsu did not even reply to the request, and just echoed the request once it expects that the unikernel is ready. This way Jitsu wouldn't even need to know the IP address of the unikernel, and maybe it could even use DHCP.

If the unikernel doesn't have its own mDNS responder then Jitsu could respond, and it could also perform the "probe" to ensure uniqueness of the hostname (if applicable).

Having Jitsu respond to DNS-SD PTR queries is another possibility.

mor1:
On 7 January 2015 at 17:28, Luke Dunstan <notifications@github.com> wrote:

> OK taking your hint, I managed to get Jitsu going with normal DNS. It can
> resume/suspend a unikernel (I don't know how to get "shutdown" to work with
> a unikernel though). I understand that Jitsu is acting as a forwarding DNS
> resolver. Is it something that you would expect to use on the internet
> directly or is it more for use on a private LAN/WAN?
>
not sure we've really worked through all possible configurations. but at
least one is jitsu-as-unikernel-inetd, ie., a running jitsu would receive
[...]

PR #35, opened 2014-12-31
mDNS changes to Trie (no SOA) and Query (multi-question, etc.)

infidel:
This is part of #28 

The changes are:

1. mDNS doesn't use SOA nor delegation (RFC 6762 section 12), so some minor changes to Trie are required to handle this.

2. mDNS doesn't echo the questions in the response (RFC 6762 section 6), except in legacy mode, so a bool argument was added to Query.response_of_answer. Perhaps there are more elegant solutions?

3. Query.answer still exists but now Query.answer_multiple is also available for answering multiple questions in one query to produce a single answer  (RFC 6762 section 5.3). One caveat is that responses may exceed the maximum message length, but that is not really specific to mDNS. Also, in theory multiple questions might require multiple separate response messages in unusual cases, but that is complicated so I am not dealing with it yet.

4. Query.answer_multiple takes an optional function to allow the caller to control the cache-flush bit. This bit is only set for records that have been "confirmed as unique". Using a callback requires minimal changes here but puts the burden of maintaining uniqueness state elsewhere. Do you think that it would be preferable to move such information into the dnsnode, even though it is mDNS-specific?

5. Query.answer_multiple takes an optional function to filter the answer, in order to support "known answer suppression" (RFC 6762 section 7.1). Again, using a callback requires minimal change to the core, but later on the mDNS-specific known answer suppression logic could move into the Query module if that turns out to be simpler.

6. A query for PTR returns additional records for SRV and TXT, to support efficient service discovery.

7. Trie.iter was added to support mDNS announcements.

8. Unit tests were added for some of the changes above. Let me know if you can see any specific gaps that need to be addressed for this PR.


avsm:
Am reviewing these patches.  There's a lot of mdns-specific stuff (e.g. the argument added to `Query.response_of_answer`, but I'm happy enough to follow the style of adding protocol-specific exceptions there for now (we have some dnssec specific bits too).  A bigger refactoring can happen once the functionality is in place and we release a 1.0.

avsm:
> Using a callback requires minimal changes here but puts the burden of maintaining uniqueness state elsewhere. Do you think that it would be preferable to move such information into the dnsnode, even though it is mDNS-specific?

Are there multiple ways a client might maintain uniqueness state (for example via Irmin synchronization so that less broadcast traffic is required)?  For now I'd keep the callback but bear in mind that if only one mechanism is used, this can be moved into `dnsnode`.

samoht:
yay! Impressive PR!

Issue #41, opened 2015-01-27
mDNS Packet.q_unicast interface confusing


dsheets:
The `Packet.q_unicast` field is used to specify that a querier is willing to receive unicast responses from a multicast message. The mDNS RFC Sec 5.4 <https://tools.ietf.org/html/rfc6762#section-5.4> says that it does this via the reserved top bit in the question class field. Usually, this field is `Q_IN` = 0x0001.

`QM` is the normal 0-top-bit value. `QU` is the special, unicast-marked packet for mDNS. If you are using `ocaml-dns` to transmit normal non-mDNS DNS, the `q_unicast` field should be (confusingly) set to `QM` otherwise the class will be unknown for classical servers (which may then refuse to forward the query or otherwise operate as expected).

dsheets:
However this gets resolved, the resolution should include documentation in the `.mli`.

infidel:
Even though the names of "q_unicast", "QM" and "QU" were chosen to match the RFC, yes it is a bit weird to have to set q_unicast=QM for non-mDNS DNS.

I guess there is no way to have a default value for a record field in OCaml? Coming from a C++/C#/Python background, I might use a constructor or other function that defaults to QM so that non-mDNS usage doesn't have to specify the value of the field at all. A constructor function is an option here, but doesn't help backward compatibility.

A simpler option would be to just rename "QM" and "QU" to something more descriptive that reflects the fact that a value of zero for the bit can mean either "non-mDNS" or "mDNS requesting multicast response" (depending on the UDP source port number). I don't have a good suggestion for these names yet.

Having three variant tags would make marshalling a bit more intuitive but the parse function would still have to one of two values for zero, so that's not ideal either.

dsheets:
I think a constructor function for the `question` type and some documentation on the question type pointing to the constructor function and describing the meaning of the `q_unicast` field would solve the problem. I've just run into an instance of use where mDNS was not being used by `q_unicast` was set to `QU` (presumably because the DNS query was not multicast).

To really clarify the situation, the constructors could be renamed from `QM` to `Q_Normal` and `QU` to `Q_mDNS_Unicast` or similar.

infidel:
OK Q_Normal and Q_mDNS_Unicast makes sense. In that case, is there still a need for a function too? (assuming that these two values are documented)

dsheets:
This was fixed by #48.

PR #48, opened 2015-01-28
Issue #41: try to make the unicast response a bit less confusing

infidel:
@dsheets Is this what you meant?

### DIFF
--- a/lib/packet.ml
+++ b/lib/packet.ml
@@ -826,12 +826,12 @@ let q_class_to_string x =
 let string_to_q_class x =
   string_to_q_class ("Q_"^x)
 
-type q_unicast = QM | QU
+type q_unicast = Q_Normal | Q_mDNS_Unicast
 
 let q_unicast_to_string x =
   match x with
-  | QM -> "QM"
-  | QU -> "QU"
+  | Q_Normal -> "Q_Normal"
+  | Q_mDNS_Unicast -> "Q_mDNS_Unicast"
 
 cstruct q {
   uint16_t typ;
@@ -845,11 +845,14 @@ type question = {
   q_unicast : q_unicast;
 }
 
+let make_question ?(q_class=Q_IN) ?(q_unicast=Q_Normal) q_type q_name =
+  { q_name; q_type; q_class; q_unicast; }
+
 let question_to_string q =
   sprintf "%s. <%s|%s%s>"
     (domain_name_to_string q.q_name)
     (q_type_to_string q.q_type) (q_class_to_string q.q_class)
-    (if q.q_unicast = QU then "|QU" else "")
+    (if q.q_unicast = Q_mDNS_Unicast then "|QU" else "")
 
 let parse_question names base buf =
   let q_name, (base,buf) = parse_name names base buf in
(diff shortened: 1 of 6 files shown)


lib/packet.ml
```diff
@@ -845,11 +845,14 @@ type question = {
   q_unicast : q_unicast;
 }
 
+let make_question ?(q_class=Q_IN) ?(q_unicast=Q_Normal) q_type q_name =
```
dsheets: Comparing `make_`, `create_`, and `new_` across the codebase, I find that `make_` and `create_` are relatively unused but the `Loader` and `Trie` modules use `new_`. These modules' types are also mutable, however, so I'm not sure if this should be `new_`. I think `make_` is probably OK.

lib/packet.mli
```diff
@@ -285,17 +285,27 @@ type q_class = Q_IN | Q_CS | Q_CH | Q_HS | Q_NONE | Q_ANY_CLS
 val q_class_to_string : q_class -> string
 val string_to_q_class : string -> q_class option
 
-(** mDNS unicast response bit. *)
-type q_unicast = QM | QU
+(** This bit is unused in normal DNS, but in mDNS a value
+    of 0 means that multicast responses are desired,
+    while a value of 1 is a request for unicast responses. *)
```
dsheets: This comment is a bit confusing because I believe that normal DNS uses the 0 value as a top bit zero in what is otherwise the query class. The comment also talks about bits but then the constructors are named.

You can also attach comments to individual constructors by putting them at the end of the constructor line (or on the next line).

lib/packet.mli
```diff
@@ -285,17 +285,27 @@ type q_class = Q_IN | Q_CS | Q_CH | Q_HS | Q_NONE | Q_ANY_CLS
...
+type q_unicast =
+  | Q_Normal
+  | Q_mDNS_Unicast
 val q_unicast_to_string : q_unicast -> string
 
-(** A question, with the usual conversion functions. *)
+(** A question, with the usual conversion functions.
+    Use make_question if you want to take advantage of default values. *)
```
dsheets: Please use `{! make_question }`.

lib/packet.mli
dsheets: You might also mention that the RFC calls these values `QM` and `QU` and add the line

```
@see <https://tools.ietf.org/html/rfc6762#section-5.4> RFC 6762 Section 5.4
```

dsheets:
This is what I meant. Looks like a good patch. See diff comments for a few miniscule nits.

infidel:
Somewhat related to RFC references, I had a thought to insert references to all of the requirements into comments in the unit test code in particular, because when implementing features piecemeal it is hard to come back later and figure out exactly what requirements are still left to do, except by reading the whole RFC each time. The problem with the comments then is that only the section headings are numbered in the RFC, so it would be necessary to have a script that splits the RFC text into sentences with unique numbering.

I have not attempted traceability to that extreme before, so it's just a thought.

dsheets:
I think referring to RFC section with ad hoc comments about missing requirements would be a huge improvement. I'm also interested in increasing the granularity of the references and I've also not attempted that level of specification document/source code before.

The patch looks good to me. If you could update `CHANGES` for `0.14.0 (trunk)` with the interface updates (changed flag constructors, added record constructor) by mimicking the previous entries that would be outstanding. I'll merge when Travis goes green.

infidel:
BTW where was the instance that you found QU being erroneously used for non-mDNS? Is it a package on OPAM?

dsheets:
MagnusS/ocaml-dnscurve@3e32e5e462f9bf65917115a71424c4a6bdf93e9c

Issue #47, opened 2015-01-28
Expose Dns_resolver.send_pkt functionality


dsheets:
Users should be able to construct custom packet to be sent rather than having to effectively use `Dns_resolver.resolve` as a packet constructor. This is probably necessary to give fine control over EDNS0 (#46) without cramming all packet construction parameters into a resolve call.

This would also benefit forwarders (which may wish to inspect/re-write a packet and then re-send).

Opinions?

avsm:
Yep, with a better than than send_pkt

dsheets:
Related to #20.

dsheets:
I think this is covered by #49. Closing.

PR #49, opened 2015-02-07
Expose Dns_resolver.resolve_pkt to support custom query packets

infidel:
Issue #47, and also needed for mDNS because queries require AA=0 and RD=0.

You might have a better suggestion for the name.


infidel:
One question: do you prefer PRs to include CHANGES to reduce effort, or does it cause merge conflicts?  (I'm not familiar with what is standard practice on GitHub)

avsm:
Either one's fine really -- fixing up a conflict in CHANGES isn't terribly difficult. It's probably easier not to include it in a pull request though.

infidel:
Is there anything missing/incorrect in this PR?

Issue #51, opened 2015-02-17
Different QNAME encoding


heidi-ann:
I've got two unikernels, one users gethostbyname and the other uses resolve. For the same query and server, the gethostbyname one works and the other doesnot (fails with no such name). On closer inspected, the only difference between the DNS query packets is the same domain name has been encoded differently in the QNAME field. I'll add an image with the two encoding in a sec

heidi-ann:
![screen shot 2015-02-17 at 17 18 16](https://cloud.githubusercontent.com/assets/1835251/6233231/5da75594-b6c9-11e4-82cc-f04a6b7dcf48.png)

dsheets:
The top packet has `google.com` as a single label (wrong) whereas the bottom packet has `google` followed by `com` (correct).

dsheets:
I think this is probably because your `resolve` call is using `["google.com"]` when it should be using `["google";"com"]`.

heidi-ann:
yes, I think thats it. thanks david :)

heidi-ann:
for future peoples ref: string_to_domain implements this

avsm:
Could you file another issue here to make this API safer?  We should protect the `resolve` call behind an abstract type that guarantees this invariant of splitting up a domain name into label chunks.

heidi-ann:
working on PR as we speak

PR #52, opened 2015-02-17
Making the type Name.domain_name abstract

heidi-ann:
This is my attempt to no longer expose that domain_name is implemented as a string list in the Name module, please review
Discussion about making domain_name abstract: https://github.com/mirage/ocaml-dns/issues/51

### DIFF
--- a/lib/name.mli
+++ b/lib/name.mli
@@ -1,6 +1,7 @@
 (*
  * Copyright (c) 2011 Richard Mortier <mort@cantab.net>
  * Copyright (c) 2011 Anil Madhavapeddy <anil@recoil.org>
+ * Copyright (c) 2015 Heidi Howard <hh360@cam.ac.uk>
  *
  * Permission to use, copy, modify, and distribute this software for any
  * purpose with or without fee is hereby granted, provided that the above
@@ -27,12 +28,27 @@ open Cstruct
 type label
 
 (** Domain name, as a list of strings. *)
-type domain_name = string list
+type domain_name
 
 (** Domain name map *)
 module Map: Map.S with type key = domain_name
 
-(** Render a {! domain_name} to a simple string. *)
+(** Create empty {! domain_name}. *)
+val empty: domain_name
+
+(** Render a {! domain_name} to a {! string} list. *)
+val to_string_list : domain_name -> string list
+
+(** Convert a standard domain {! string} list to a {! domain_name}. *)
+val of_string_list : string list -> domain_name
+
+(** Append two {! domain_name} values. *)
+val append: domain_name -> domain_name -> domain_name
+
+(** Concatenate a {! string} onto a {! domain_name}. *) 
+val cons: string -> domain_name -> domain_name
+
+(** Render a {! domain_name} to a simple {! string}. *)
 val domain_name_to_string : domain_name -> string
 
 (** Convert a standard domain {! string} to a {! domain_name}. *)
@@ -74,3 +90,4 @@ type key = string
 val canon2key : domain_name -> key
 
 val dnssec_compare : domain_name -> domain_name -> int
+val dnssec_compare_str : string list -> string list -> int
--- a/lib/zone_parser.mly
+++ b/lib/zone_parser.mly
@@ -362,19 +362,19 @@ hexword: charstring
    assume the owner field was omitted */
 owner:
    /* nothing */ { state.owner }
- | domain { state.owner <- $1 ; $1 }
+ | domain { state.owner <- $1 ; state.owner }
 
 domain:
-   DOT { [] }
+   DOT { Name.empty }
  | AT { state.origin }
- | label_except_at { (Bytes.lowercase $1) :: state.origin }
- | label DOT { [ (Bytes.lowercase $1) ] }
- | label DOT domain_labels { ((Bytes.lowercase $1) :: $3) @ state.origin }
- | label DOT domain_labels DOT { (Bytes.lowercase $1) :: $3 }
+ | label_except_at { Name.cons $1 state.origin }
+ | label DOT { Name.of_string_list ([$1]) }
+ | label DOT domain_labels { Name.cons $1 (Name.append (Name.of_string_list $3) state.origin) }
+ | label DOT domain_labels DOT { Name.of_string_list ($1 :: $3) }
 
 domain_labels:
-   label { [(Bytes.lowercase $1)] }
- | domain_labels DOT label { $1 @ [(Bytes.lowercase $3)] }
+   label { [$1] }
+ | domain_labels DOT label { $1 @ [$3] }
 
 /* It's acceptable to re-use numbers and keywords as character-strings.
    This is pretty ugly: we need special cases to distinguish a domain
--- a/lib/loader.ml
+++ b/lib/loader.ml
@@ -366,9 +366,9 @@ type parserstate = {
   mutable paren: int;
   mutable filename: string;
   mutable lineno: int;
-  mutable origin: string list;
+  mutable origin: domain_name;
   mutable ttl: int32;
-  mutable owner: string list;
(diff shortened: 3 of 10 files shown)


lib/name.mli
```diff
@@ -27,11 +27,20 @@ open Cstruct
 type label
 
 (** Domain name, as a list of strings. *)
-type domain_name = string list
+type domain_name
```
dsheets: Would be nice to just call this `t` as we're breaking the interface anyway. `Name.domain_name` is a bit verbose.

lib/name.mli
```diff
@@ -27,11 +27,20 @@ open Cstruct
...
 
 (** Domain name map *)
 module Map: Map.S with type key = domain_name
 
+(** Create null domain name *)
+val empty: domain_name
```
dsheets: Prefer "empty" to "null" in doc. Could also call this object "root" as it's the zero-label list (parent of all names).

lib/name.mli
```diff
@@ -27,11 +27,20 @@ open Cstruct
...
+
+(** Render a {! domain_name} to a {! string} list. *)
+val domain_name_to_string_list : domain_name -> string list
```
dsheets: Prefer "to_string_list" and comment like "[to_string_list name] is the list of labels in the name [name]"

lib/name.mli
```diff
@@ -27,11 +27,20 @@ open Cstruct
...
+
+(** Convert a standard domain {! string} list to a {! domain_name}. *)
+val string_to_domain_name : string list -> domain_name
```
dsheets: Prefer "of_string_list" and same doc style as above.

lib/name.mli
```diff
@@ -27,11 +27,20 @@ open Cstruct
...
+
 (** Render a {! domain_name} to a simple string. *)
 val domain_name_to_string : domain_name -> string
```
dsheets: You should probably break these as well to make them match with the above style which is much preferred over the present style (imo).

dsheets:
I think this is the right approach. This whole interface could be improved if we're going to break compat by hiding the type alias to `string list`. I'd prefer to standardize the style of the interface and break as much as possible in a single release than dribble out changes and live with weird names forever.

dsheets:
Also, the `open Cstruct` at the beginning is annoying (makes `t` mean `Cstruct.t` when reading. If you update the interface, please add yourself to the file header.

lib/name.mli
heidi-ann: i agree, will try to get this compiling, then do a big find replace

lib/name.mli
heidi-ann: called in null  185ce2f5f3bbd7f1218ac62ed396e687d2b2cc09

dsheets: Oops, I meant in the documentation string "null" is a bit confusing. Does it mean using it is invalid? The word "empty" or "zero" or "root" is more descriptive, IMO.

heidi-ann:
So, I can't seem to get rid of the following type error, despite casting all the string lists to  Name.domain_name 

```
File "lib/zone_parser.mly", line 227, characters 70-72:
Error: This expression has type string list
       but an expression was expected of type Name.domain_name
Command exited with code 2.
```

dsheets:
https://github.com/mirage/ocaml-dns/blob/master/lib/loader.mli#L89

heidi-ann:
should https://github.com/mirage/ocaml-dns/blob/master/lib/loader.mli#L91 be a domain_name too?

heidi-ann:
im stalled on
```
File "lib/zone_parser.mly", line 364, characters 19-30:
Error: This expression has type string list
       but an expression was expected of type Name.domain_name
```
the types seem to propagate in strange ways in the yacc file

dsheets:
https://github.com/mirage/ocaml-dns/blob/master/lib/loader.mli#L38 looks like the RR adders take `domain_name` for `owner` but `state.owner` is `string list`.

heidi-ann:
thanks david, i'll switch owner to domain_name after the meeting

lib/name.mli
heidi-ann: i'd prefer to do this change after this PR, otherwise it will pollute the diff files, making them difficult to read

lib/name.mli
heidi-ann: fixed, now called empty in function name and doc

lib/name.mli
heidi-ann: now called to_string_list, i agree with preference for [ ] to {! } but I'll leave for another PR as this is consistent throughout atm

lib/name.mli
heidi-ann: i naively tried this, noticed how pervasive ``open Cstruct`` was gave up :)
commit 17f69f5fdc155b878d03111cbc4353c31fdf06c7
revert 1593fa76b606405a7842dffe693e2e169f1fa9f5

lib/zone_parser.mly
```diff
@@ -362,15 +362,15 @@ hexword: charstring
...
+   DOT { Name.empty }
  | AT { state.origin }
- | label_except_at { (Bytes.lowercase $1) :: state.origin }
- | label DOT { [ (Bytes.lowercase $1) ] }
- | label DOT domain_labels { ((Bytes.lowercase $1) :: $3) @ state.origin }
- | label DOT domain_labels DOT { (Bytes.lowercase $1) :: $3 }
+ | label_except_at { Name.cons (Bytes.lowercase $1) state.origin }
+ | label DOT { Name.of_string_list ([Bytes.lowercase $1]) }
+ | label DOT domain_labels { Name.cons (Bytes.lowercase $1) (Name.append (Name.of_string_list $3) state.origin) }
+ | label DOT domain_labels DOT { Name.of_string_list ((Bytes.lowercase $1) :: $3) }
```
dsheets: No need for `Bytes.lowercase` here any more.

heidi-ann: done bef3460b62353471572ae9509400ec8b85de52ba

heidi-ann:
@dsheets tada! now as one commit

lib/zone_parser.mly
```diff
@@ -362,19 +362,19 @@ hexword: charstring
...
+   DOT { Name.empty }
  | AT { state.origin }
- | label_except_at { (Bytes.lowercase $1) :: state.origin }
- | label DOT { [ (Bytes.lowercase $1) ] }
- | label DOT domain_labels { ((Bytes.lowercase $1) :: $3) @ state.origin }
- | label DOT domain_labels DOT { (Bytes.lowercase $1) :: $3 }
+ | label_except_at { Name.cons $1 state.origin }
+ | label DOT { Name.of_string_list ([$1]) }
+ | label DOT domain_labels { Name.cons $1 (Name.append (Name.of_string_list $3) state.origin) }
+ | label DOT domain_labels DOT { Name.of_string_list $1 :: $3) }
```
dsheets: Syntax error here.

Also, could you please make the commit message more descriptive? Something like "Make Name.domain_name abstract" maybe?

samoht: > Syntax error here.

who needs a compiler when you have @dsheets to automatically compile your PR in his head? :p

heidi-ann: syntax errors fixed and commit msg edited

infidel:
When I checked out your PR and then typed "make" it didn't compile:

```
File "lib_test/unix/time_server.ml", line 28, characters 6-10:
Error: This expression has type string list
       but an expression was expected of type Dns.Name.domain_name
Command exited with code 2.
```

But yes, #54 needs to be fixed too.

heidi-ann:
@infidel thanks, I forgot to test on unix as well as xen backends. I'll fix soon

heidi-ann:
Sorry its taken me so long, I believe its working now

PR #56, opened 2015-03-03
Minimal mDNS responder and resolver

infidel:
This should be enough to start the discussion about interface design. I think that the responder is useful for non-unique records and for the simplest unique record: a single hostname that maps to an A record. Supporting more complex cases of unique records seems to be fairly involved (lexicographical comparison for tiebreaking).

The resolver is a lot simpler (minimal), apart from the unit tests, which use somewhat verbose mock modules (kind of as an experiment to see how testable dns.mirage is). Efficiency improvements such as caching, including known answers in queries, and so on would be nice to have later.

The main part of https://github.com/infidel/ocaml-mdns that will have to wait for a later PR is the acceptance test scripts.

Reference: https://tools.ietf.org/html/rfc6762


dsheets:
This looks really nice! I'm prepping a merge to take into account #52 and #64 which I will PR to @infidel.

I haven't combed the source but it looks very reasonable and readable. My only immediate observations about the interface is:

I don't see a way to get the claimed name back from the responder after attempting to load a unique name. Perhaps `add_unique_name` should return a `Name.t Lwt.t` to be fulfilled after the probe completes? If I read the RFC right, it's not typically possible to lose a name once you have it unless you fail to defend within 750ms... I wonder if we should have a channel for information about name loss as well?

I'll send a PR for the merge and take a closer look tomorrow. Thanks so much for this! It's very exciting and I apologize that it's taken us so long to get around to looking at this.

Out of curiosity, do you have any immediate applications you'd like to use this for? I'm keen on a Bonjour Sleep Proxy for WoL of a Mac Mini... and of course random Jitsu/Mirage services. :-)

infidel:
Thanks for the comments. I was worried that the style might be too C++-ish. I like the idea of the deferred for the final unique name. As you say it is very unlikely to lose a unique name, so having a channel for that specifically could be more of a wish list item.

I will probably have to make the first change next week when I am back home from a business trip.

dsheets:
I've sent you a PR to this branch to bring it up-to-date wrt the `Name` interface modifications. `Name` might still change slightly (`canon2key` is a horrible name).

I've also left inline comments about the mDNS responder source. Overall, I think we should merge this as soon as it's feasible and then start filing issues in the ocaml-dns tracker on all missing components/features.

Right now, my two largest concerns about the implementation are lexicographic comparison while masking cache flush bit and termination conditions. I made a couple comments about 'goodbye' messages with TTL 0.

dbuenzli:
Le mardi, 31 mars 2015 à 12:09, David Sheets a écrit :
> canon2key is a horrible name

Yes ! When I was looking your `Name` patch yesterday this was constantly catching my eye but I decided not to comment on (not enough into the subject to have a good suggestion to make and sometimes I feel that I'm constantly annoying people on names so I try to refrain).  

Daniel

avsm:
Your naming suggestions are always useful, irrespective of their annoyance factor, so please feel free to keep them up :-)

dsheets:
It's been changed to `to_key` in 548f359 for infidel/ocaml-dns#1.

infidel:
The couple of build failures here look like they are unrelated to this PR. Do you agree?

I thought about the idea of returning the possibly-renamed unique hostname after probing, but I would like to leave it until a separate issue/PR.

Have the commit comments been sufficiently addressed yet?

dsheets:
Indeed the build failures are unrelated. Thanks for doing all the work to make mDNS possible in ocaml-dns!

PR #58, opened 2015-03-18
adding process composition function to Dns_server

heidi-ann:
I thought this function might be useful for other projects too

### DIFF
--- a/lwt/dns_server.mli
+++ b/lwt/dns_server.mli
@@ -33,6 +33,9 @@ end
 
 type 'a processor = (module PROCESSOR with type context = 'a)
 
+(** [compose process backup_process] takes two packet processes, returning one
+	which applies the second process only the first returns ServFail or NXDomain *)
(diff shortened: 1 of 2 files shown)


lwt/dns_server.mli
```diff
@@ -33,6 +33,9 @@ end
 
 type 'a processor = (module PROCESSOR with type context = 'a)
 
+(** [compose process backup_process] takes two packet processes, returning one
+	which applies the second process only the first returns ServFail or NXDomain *)
+val compose: Dns.Packet.t process -> Dns.Packet.t process -> Dns.Packet.t process
```
dbuenzli: I think the doc string could be improved. It's usually better to describe the result, in this case the resulting process, in term of the variables you mention. IIUC what this does I would tentatively write it this way: 
```ocaml
(** [compose process backup_process] is [process] unless it returns {!ServFail} or {!NXDomain} in which case it becomes [backup_process]. *)
```

PR #61, opened 2015-03-20
Separating pure string operations from mldig

heidi-ann:
This PR makes the following changes:
- renaming mldig.ml to mldig_unix.ml due to use of Unix
- creates a new mldig.ml and mldig.mli for the pure operations
- adds the string_of_x function to mldig.ml

The purpose of this change is allow mirage application to use digs pretty printing of Dns packets.


dsheets:
Could we keep the tool named `mldig` but name the module just `Dig`? Otherwise, this looks good.

Also, could you modify the Travis script to stop calling `make dns-depend` since <https://github.com/mirage/mirage-skeleton/commit/93cc9954666c859f6afe552ebcaafc9a70af1ec8>?

heidi-ann:
how can i tell oasis to name the executable mldig.native instead of dig_unix.native

currently the oasis file says:
```
Executable mldig
  Path:               lwt
  MainIs:             dig_unix.ml
  Build$:             flag(tests) && flag(lwt)
  Custom:             true
  CompiledObject:     best
  Install:            false
  BuildDepends:       lwt, lwt.unix, lwt.syntax, dns, dns.lwt, re, re.str, ipaddr, cmdliner
```

avsm:
The binary will always be built as its filename, but oasis will install it as `mldig`.  Why have you renamed it to `dig` though?  It should not be confused with the traditional `dig` command installed by BIND.

avsm:
(The splitting of Mldig's functionality so that Mirage apps can use it is great)

heidi-ann:
@dsheets asked me to rename the modules from mldig to dig

There's more work to done before theres a mirage dig but i just wanted the pretty printing functionality to debug mirage apps

dsheets:
Sorry, I wasn't clear. I meant naming the module containing the `dig` functionality `Dig` but the executable front-ends named `mldig`.

dsheets:
the module containing the *generic* `dig` functionality

heidi-ann:
I think we are all in agreement. We want the modules called dig but the executable mldig to avoid a naming collision with normal dig

avsm:
Has this been merged separately, @dsheets?

dsheets:
Nope, I thought there was some issue with the executable name... but it looks OK to me right now?

PR #64, opened 2015-03-30
Rationalize and encapsulate Name

dsheets:
This patchset attempts to rationalize the interface `Name` by incorporating @heidi-ann's #52 and then renaming many of the exposed values in `Name` to make them shorter while discouraging the `open`ing of `Name`. I also rewrote some of the documentation in `Name` and updated all the internal dependencies.

`Name.for_reverse` also changed signature and now can create `PTR` names for IPv4 or IPv6 addresses. This requires a minimum version of `Ipaddr` of 2.6.0.

I also did some general clean-up of `open Cstruct` as `Cstruct` has a type `t`.

I would really appreciate review of the new `Name` interface as I'm keen to break it as few times as possible. Thanks!


dsheets:
There appear to be 3 reverse dependencies:

* mirage-dns which builds fine
* alpaca which requires `base64 < 2.0.0` and so is incompatible because we require `base64 >= 2.0.0`
* libres3 which fails due to `Dns.Name.string_to_domain_name`

CHANGES
```diff
@@ -1,3 +1,13 @@
+0.15.0 (trunk):
+* Name.domain_name has been renamed to Name.t and is now abstract
+* Name.domain_name_to_string has been renamed to Name.to_string
```
avsm: worth keeping domain_name_to_string as an alias to keep libres3 working for a while?
https://travis-ci.org/mirage/ocaml-dns/jobs/56403232#L2134

lib/name.mli
```diff
@@ -21,43 +23,62 @@
...
+    serialization of [name] without the trailing dot. *)
+val to_string : t -> string
 
-(** Parse a {! domain_name} out of a {! Cstruct.t} given a set of already
+(** [of_string name] is the domain name parsed out of the normal,
+    human-readable serialization [name]. *)
+val of_string : string -> t
+
+(** [for_reverse ip] is the name used for reverse lookup of IP address [ip]. *)
+val for_reverse : Ipaddr.t -> t
```
avsm: why not just call this `of_ipaddr`?  The `for_reverse` is a bit confusing until you examine the type more closely.

avsm:
Lovely changeset.  How does it merge with #56?

CHANGES
dsheets: `string_to_domain_name` has been retained but deprecated in new commit.

lib/name.mli
dsheets: Totally agree. It's changed now.

dsheets:
I'm still doing background reading for #56. I will review it and any conflicts as my next priority.

PR #72, opened 2015-06-05
Add many attributes referencing the relevant parts of RFC 6762

infidel:
Part of #28. I can add the annotated RFC (XML) to this PR if the attributes are acceptable. I plan to add it in a "doc" subdirectory unless someone has a better idea.


hannesm:
any status on this?

yomimono:
Ah, I just closed the referenced issue thinking that the work in `lwt/mdns_responder.ml` etc implemented what was discussed there.  I'm going to close this PR as well, but please feel free to update us on the status if you're still working on or thinking about this.

PR #79, opened 2015-07-03
Remove dep to `mirage-tcpip`, remove camlp4 when it's not needed

samoht:


samoht:
@dsheets what do you think?

_oasis
```diff
@@ -65,8 +62,7 @@ Library "dns-lwt-core"
   Install$: flag(lwt)
   Findlibname: lwt-core
   Modules: Dns_server, Dns_resolver, Mdns_responder
-  BuildDepends: lwt, cstruct, lwt.syntax, dns, mirage-profile
-  XMetaRequires: lwt, cstruct, dns, mirage-profile
+  BuildDepends: lwt, cstruct, lwt, dns, mirage-profile
```
dsheets: two `lwt`s

lwt/dns_resolver.ml
```diff
@@ -48,40 +48,37 @@ let send_pkt ?alloc client ({ txfn; rxfn; timerfn; cleanfn }) pkt =
...
-    >>= fun r ->
-    commfn.cleanfn ()
-    >>= fun () ->
-    return r
-  with exn ->
-    commfn.cleanfn ()
-    >>= fun () ->
-    fail exn
+  Lwt.catch (fun () ->
+      send_pkt ?alloc client commfn pkt >>= fun r ->
```
dsheets: I like the removal of the syntax extension but I dislike binding at the *end* of the line.

lwt/dns_resolver_unix.ml
```diff
@@ -123,8 +124,8 @@ module Resolv_conf = struct
     )
 
   let create client ?(file=default_file) () =
-    lwt t = get_resolvers ~file () in
-    return {
+    get_resolvers ~file () >>= fun t ->
```
dsheets: EOL binding and could use `>|=` to avoid the noisy `Lwt.return`

lwt/dns_server.ml
```diff
@@ -40,22 +40,22 @@ let compose process backup ~src ~dst packet =
...
   | None -> backup ~src ~dst packet
 
 let process_query buf len obuf src dst processor =
   let module Processor = (val processor : PROCESSOR) in
   match Processor.parse (Dns.Buf.sub buf 0 len) with
-  |None -> return None
+  |None -> Lwt.return None
   |Some ctxt -> begin
-    lwt answer = Processor.process ~src ~dst ctxt in
+    Processor.process ~src ~dst ctxt >>= fun answer ->
```
dsheets: EOL binding and could use `>|=`

dsheets:
Overall, I think this improves the package and I'd love to merge it when the nits are cleared up.

lwt/dns_resolver.ml
samoht: only here or everywhere?

dsheets: In general, I think it's confusing and easy to miss bindings. I can see the case for `>>= fun () ->` at the EOL but introducing new names on the RHS of the text always feel awkward to me.

I guess the style I'd like for the whole codebase is `>>= fun _ ->` at the beginning of line or, if the body is long, on a line of its own.

samoht:
ok, I've pushed some changes. I'm not very fond of trading horizontal space for vertical space, but that's ok I guess :-)

dsheets:
I agree that it's better generally to use horizontal space than vertical space unless it makes reading harder/slower.

Also, I have this sweet dual portrait setup so I can have 132 lines of a buffer viewable at a time...

samoht:
ok, I've changed all occurrences of `>>= fun _ ->` to be at the end of line only if `_` is not `()`. Should be good enough to merge I think.

dsheets:
LGTM. I'll wait for the green.

samoht:
Actually, I have other minor cleanups, I'll push them soon.

dsheets:
Hooray! Could you also set the required lwt to `> 2.4.7` in `_oasis` and `opam` for `Lwt.Infix`?

dsheets:
Also, why did <https://github.com/mirage/ocaml-dns/commit/7612f6b0e078f54c8e97bd2952a79a1ade09108d> happen?

samoht:
it's green!

samoht:
I've also adde mirage-dns.opam and removed the dependency to `tcpip` here as well. This will need a new release of both packages.

PR #80, opened 2015-07-30
Answer more than one query in dns_server_unix.listen

MagnusS:
There is a regression in 0.15.2 that results in only the first DNS query being answered when using `Dns_server_unix.listen` (the loop [here](https://github.com/mirage/ocaml-dns/blob/master/lwt/dns_server_unix.ml#L81) is only called once). This PR should fix that.


dsheets:
That's a pretty severe regression. Thanks! Under publication review as ocaml/opam-repository#4565.

PR #82, opened 2015-08-13
mDNS refactoring and misc fixes

infidel:
The primary change here is to factor out the state machine for mDNS probing into a new module Dns.Probe, which is almost purely functional (apart from the trie when renaming), inspired a little by ocaml-tls. Hopefully it is easier to understand and more reusable for the future (e.g. with Async). There is still a fair bit of logic in Mdns_responder that could further be moved to a functional layer (named Dns.Multicast perhaps?). This change doesn't affect functionality or the public interface of Mdns_responder.

Some additional small changes in this PR:
- Fixes a bug in the handling of identical mDNS probes that I noticed while refactoring, and adds a test.
- Added a functor Mdns_resolver_mirage.Chain that is intended to compose an mDNS resolver with a normal DNS resolver such that *.local is resolved via mDNS and everything else is done with DNS.
- Changed Dns.Query to not respond to queries for classes other than IN.
- Fixed mDNS legacy responses to use TTL <= 10 sec.
- Fixed mDNS responses to use RD=0.

This PR would conflict with #72, so I'll fix that one some other time.


lib/probe.ml
```diff
@@ -0,0 +1,352 @@
...
+   - Event: send complete
+   - (To Do: send fail or network down? trie changed?)
+
+   Outputs (actions):
+   - Packets to be sent
+   - Delay time
+   - Idle/finished indications
+*)
+
+let multicast_ip = Ipaddr.V4.of_string_exn "224.0.0.251"
```
avsm: this might be obvious, but any chance of making this `multicast_dns_ip`?  I had to look it up to make sure this was the case...

lib/probe.ml
```diff
@@ -0,0 +1,352 @@
...
+      ({
+        state with
+        stage = SendingProbe { datagram; num=FirstProbe; rrs };
+        names_pending = UniqueSet.empty;
+        names_probing = state.names_pending;
+      }, ToSend datagram)
+  in
+  match state.stage with
+  | ProbeIdle
+    ->
```
avsm: odd formatting here -- could the `->` go after the `ProbeIdle`?

lib/probe.ml
```diff
@@ -0,0 +1,352 @@
...
+  | DelayBeforeRestart
+  | ProbeStopped ->
+    (* No change *)
+    state
+
+let on_send_complete state =
+  match state.stage with
+  | SendingProbe probing
+    ->
+    (* Fixed delay of 250 ms *)
```
avsm: Do these timeouts ever need to be changed (e.g. in a test harness)? Or is the intention that the callback that receives `Delay` can change them?  I'm just wondering if the 250ms value should be an optional argument to `on_send_complete` rather than hardcoded here.

avsm:
This is awesome.  I'd quite like to get a DNS 1.0.0 out, and then make a subsequent version 4.02.0+ only (thus letting #72 aso be merged).

lib/probe.ml
infidel: OK, if you think it's more readable

lib/probe.ml
infidel: OK

lib/probe.ml
infidel: The spec says 250ms: http://infidel.github.io/ocaml-mdns/rfc6762_notes.html#s8.1_p4_c1

In my unit test (lib_test/ounit/test_mdns_responder.ml) I verify that the delay is correct but don't actually wait for real time to pass, so yes, the callback is intended to allow that.

infidel:
I agree about #72 waiting. Are there any particular features/qualities that you think are required to call it 1.0?

infidel:
The Travis failure here was some transient problem I guess?

avsm:
It's the Irmin.unix problem.  This should be mergeable now... will take a look shortly.

PR #83, opened 2015-08-14
change source port randomization to avoid overflow

yomimono:
Avoid choosing source ports greater than 65535, as S.listen_udpv4 does not currently detect or fix the overflow (see https://github.com/mirage/mirage-tcpip/pull/167 ).

It's not clear to me what the original logic was intended to reflect (perhaps it used to be 65300 + 255?); if this fix is also wrong, I'm happy to change it.


dsheets:
This fix looks correct to me but @avsm may be able to speak to the previous choice.

avsm:
Original logic was indeed +255 that got mysteriously morphed to +1024.  Fix looks correct to me!

PR #86, opened 2016-03-11
Remove fork of hashcons, clarify LICENSE

djs55:
This library now depends on `hashcons` rather than containing a fork of it. Now that there is no LGPL (+ linking exception) code left, clarify that the license is ISC. Previously the `opam` file claimed a mixture of licenses.


djs55:
One of the travis jobs failed with
```
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).

The job has been terminated
```

avsm:
The log length is due an upgrade/downgrade cycle of Core to jump past the sexplib bump. We'll have to move past that soon, so this library is about to become 4.02+ only...

PR #89, opened 2016-03-12
Switch to ppx, remove use of camlp4

djs55:
This requires cstruct >=1.9.0 (NB 1.9.0 supports both camlp4 and ppx)

See [mirage/ocaml-cstruct#95]


djs55:
CentOS 6 and CentOS 7 failed with:
```
package xen-devel is not installed
opam-depext: internal error, uncaught exception:
             Failure("OS package installation failed")
```
Alpine failed with:
```
# src/bigstring_stubs.c:36:24: fatal error: sys/endian.h: No such file or directory
# compilation terminated.
# Command exited with code 2.
```

avsm:
Alpine failure is fixed in latest core

djs55:
OK then, so I think all the CI failures are external:

- CentOS failures are due to jitsu not being compatible with CentOS (because of its poor xen support)
- alpine failure is fixed elsewhere

So I think we could merge this with a clear conscience :)

samoht:
MERGE IT!

djs55:
OK!

PR #90, opened 2016-03-12
Prepare to release 0.18.0

djs55:
This version uses ppx rather than camlp4.


djs55:
The failures are the known jitsu revdeps failures on CentOS 6, 7 and the known alpine failure.

PR #92, opened 2016-04-17
Query.create: do not set AA on queries

dsheets:
Some DNS servers (e.g. the one in the Technicolor TG582n) will check this bit is unset and ignore packets if it is set.

This issue has been present since this function was written in 2012.


djs55:
travis seems to have hit [gregtatcam/imaplet-lwt#11]

djs55:
Looks safe to merge anyway.

PR #94, opened 2016-06-16
stub implement pseudoheader, now required by mirage-types

hannesm:
@yomimono this makes the tests compile (for some reason the Dns_resolver_unix::resolution-error test fails on my FreeBSD)

### DIFF
--- a/lib_test/ounit/test_mdns_resolver_mirage.ml
+++ b/lib_test/ounit/test_mdns_resolver_mirage.ml
@@ -79,6 +79,9 @@ module StubIpv4 (*: V1_LWT.IPV4 with type ethif = unit*) = struct
   let checksum buf bufl =
     0
 
+  let pseudoheader t ~dst ~proto len =
+    Cstruct.create 0
+
   let get_source t ~dst:_ =
     t.ip
 


hannesm:
nearly all builds completed successfully; only some alpine-3.3 couldn't compile lwt...

samoht:
yes the alpine error is because they lwt configuration script needs ncurse which is not installed by default there ... @avsm and I tracked that issue yesterday, should be fixed soon.

yomimono:
Thanks for this @hannesm :) I've restarted the tests; we'll see how they do.

hannesm:
there was only a single failing build, you can restart individual builds... all others were fine

yomimono:
Ah, thanks, I didn't realize that - barely noticed the "restart" button on the main page, let alone within the individual builds.  I'll waste less compute cycles next time.

hannesm:
(and the alpine build is unlikely to be working before something like https://github.com/ocaml/opam-repository/pull/6791 gets merged)

hannesm:
same as yesterday, IMHO this can safely be merged since all other platforms are happy (and it fixes make test with mirage-types master).

hannesm:
merged this minor fix (won't fix your alpine opam troubles, but there are several issues in the opam repository)

PR #98, opened 2016-09-18
Preliminary TCP client support

djs55:
DNS queries and responses should be sent over TCP in certain circumstances (e.g. if the message is too big for a datagram). This PR adds the most basic possible TCP support to the `Dns_resolver_unix` module and the `dig` command-line tool. It is now possible to force the use of TCP or UDP as follows:

```
./dig_unix.native @192.168.1.2 www.google.com --proto tcp
```

By default the resolver uses UDP as before.

### DIFF
--- a/lwt/dig_unix.ml
+++ b/lwt/dig_unix.ml
@@ -35,7 +35,7 @@ open Cmdliner
 
 let dns_port = 53
 
-let dig source_ip opt_dest_port q_class q_type args =
+let dig source_ip opt_dest_port q_class q_type _proto args =
   Dns_resolver_unix.create ()
   >>= fun res ->
   let timeout = 5 (* matches dig *) in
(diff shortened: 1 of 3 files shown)


hannesm:
as a side note, when I thought about DNS over TCP, please keep in mind: a) keep sessions alive, otherwise you end up in poor performance (full 3way-handshake just for a name lookup; and b) replies may be delivered out of order, request A B C might be answered in B C A -- deliver the replies to the waiting application out of order as well (otherwise everything is waiting for A)

djs55:
@hannesm thanks for the suggestions! I'm playing with some of these ideas over in [djs55/ocaml-djs-forwarder]. When I get something that looks good I'll try porting them across to here.

avsm:
Will not fix this in this version of ocaml-dns

PR #116, opened 2016-12-22
Adapt to the new error scheme

samoht:


samoht:
Revdep error should be fixed by https://github.com/djs55/ocaml-dns-forward/pull/50

Issue #106, opened 2016-11-02
could mirage tests be using mirage-vnetif?


yomimono:
The mock network modules in `lib_test/ounit/test_mdns_responder_mirage` have a lot of boilerplate and stubbed-out code.  Could they be replaced by the [mirage-vnetif](https://github.com/magnuss/mirage-vnetif) library, which is intended to provide in-memory network stacks for testing?

avsm:
Since #159 is merged, we don't have any test cases for the mirage- or lwt-specific backends (but extensive tests for the core suite).  Leaving this open for a mirage-vnetif test suite for this library

hannesm:
when adding mirage-vnetif based tests, please investigate first what you like to test... the mirage layer is rather slim (mainly tcp connection management), most of the server and resolver logic can be tested in a pure fashion -- no need for vnetif :)

PR #122, opened 2017-01-07
refer to mirage module types modules by new name

yomimono:
V1 is now Mirage_types, and V1_LWT is now Mirage_types_lwt, as of MirageOS version 3.0.0.


yomimono:
These changes are all in commented-out annotations; I'm going to remove the comments instead, but note also that the annotations are themselves to the modules mentioned in #106 .

PR #124, opened 2017-01-11
fix typo mirage-protocol_s_

hannesm:


avsm:
good catch!

hannesm:
@avsm I click on all the red buttons of Travis and your CI... it is great how many issues they raise

PR #128, opened 2017-02-12
remove Buf module, now that cstruct is mature

hannesm:
@dsheets told me when `Buf` was written, cstruct was not yet in a mature state.  There was no performance measurement of buf vs cstruct done.

Since I like less code more, this seemed to be a good excuse to read through ocaml-dns ;)

any opinions /cc @avsm @djs55 @dsheets @yomimono


lib/protocol.ml
```diff
@@ -36,8 +36,8 @@ module Client : CLIENT = struct
     Random.self_init ();
     Random.int (1 lsl 16)
 
-  let marshal ?alloc q =
-    [q.Packet.id, Packet.marshal (Buf.create ?alloc 4096) q]
+  let marshal q =
+    [q.Packet.id, Packet.marshal (Cstruct.create 4096) q]
```
dsheets: What is `?alloc` removed here? Is it no longer necessary for page-aligned allocations in Mirage on Xen?

hannesm: good remark.  at the moment (see https://github.com/mirage/mirage-net-xen/pull/17), Mirage on Xen (in mirage-net-xen) copies the frame anyways to (half-page) aligned memory.  no need for page aligned on kvm afaik.

this allows us to remove the entire allocation magic from ocaml-dns, and use `Cstruct.create` instead.  I partially started this (by removing `?alloc` parameters), but didn't finish (will do so soon).  no io-page will be harmed by this change ;)

hannesm:
superseeded by #132

PR #130, opened 2017-02-15
Prepare to release v0.19.1 again

djs55:
Previously a v0.19.1 release was started but never tagged. Let's
try again but including the DNS parsing pointer to pointer fix.

Signed-off-by: David Scott <dave@recoil.org>

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,9 @@
-## 0.19.1 (2017-02-09)
+## 0.19.1 (2017-02-15)
 
 * Use topkg instead of oasis (#126)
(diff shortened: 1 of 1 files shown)


pkg/META
```diff
@@ -1,4 +1,4 @@
-version = "%%VERSION_NUM%%"
+version = "0.19.1"
```
hannesm: this shouldn't be done in git.. if you use `topkg distrib` and `topkg opam pkg`, the watermarks are replaced in the resulting tarball... no need to change in the repository

PR #132, opened 2017-02-24
No alloc

hannesm:
rebased #128 on master, and removed the allocation interface.  see https://github.com/mirage/mirage-net-xen/pull/17 for reasoning why page-aligned allocation is not needed.

the allocation strategy is modified (API is changed), this may also lead to different memory behaviour.


hannesm:
I'm happy to hear your opinions (@dsheets @avsm @djs55) [and/or performance analysis of dns-forward, which atm doesn't seem to build in a MirageOS3 universe] -- I'm happy to update the mirage-skeleton dns example

hannesm:
this is now lying around since two weeks, anyone had a chance to look at it, and would like to share their opinion?  this is a blocker (/precondition) for other refactorings I've in preparation, such as not throwing exceptions.

dsheets:
LGTM

lib/packet.ml
```diff
@@ -1489,12 +1488,12 @@ let parse buf =
   (* eprintf "RX: %s\n%!" (to_string dns); *)
   dns
 
-let marshal txbuf dns =
+let marshal dns =
+  let txbuf = Cstruct.create 4096 in
```
avsm: It's this `Cstruct.create` that could potentially cripple performance for a caller, since there is now choice but to allocate a page for every packet. Previously, the caller of the interface would control memory allocation and could reuse pages from a free pool if it was safe.

Why not just make txbuf an optional function that can be passed in, that does the allocation (defaulting to Cstruct.create)?

OCaml doesn't deal particularly well with rapid page allocation via Bigarrays, since there is also a C-level proxy object that (unnecessarily in our use case) reference counts the memory, so I'm reluctant to make ocaml-dns force memory allocation on the caller unless there's a really good reason.

hannesm:
@avsm I re-added a `?alloc:(unit -> Cstruct.t)` argument to `Packet.marshal`.  is this good to go now?

avsm:
lgtm; @djs55 is having a review also and then will merge/release.

djs55:
LGTM also.

PR #134, opened 2017-03-23
release 0.20.0

avsm:


opam
```diff
@@ -46,7 +46,7 @@ depends: [
   "lwt"              {test}
   "ounit"            {test}
   "pcap-format"      {test}
-  "mirage-protocols" {test}
+  "mirage-protocols" {test & >="1.1.0"}
```
hannesm: this is not needed (I initially had this as part of 132, but removed it): ocaml-dns tests don't ever call `mtu`, but provide `mtu` (for compatibility with mirage-protocols-1.1.0).  but this constraint here also does not hurt too much, so we can leave it here.

avsm: ok will leave it -- might as well test with the latest revision.

Issue #137, opened 2017-04-20
Name.ml invariants


hannesm:
does anyone feel responsible for this library?  I spend last evening reading through `name.ml`, and was wondering about the design decisions made here:
- `type t` is abstract (this is great!)
- `to_key` does some domain name validation (`0 < length < 64` and each character must be non-zero)
- both `of_string` and `of_string_list` (and cons) apply `String.lowercase` (but do no other checks)
- `parse` (rather `parse_label`) does not validate any label content (but by construction restricts size to be correct (0 < length < 64))
- exceptions are thrown all over

Since `t` is abstract, this allows us to specify invariants and check them for construction -- there will never be a `t` which violates these (they could contain the label length, and which ASCII characters may be within a label).  But AFAICS, there are none maintained.

Both `Name.Map` and `Name.Set` use _case-sensitive_ comparison, which is not very convenient if you actually want to use them (DNS labels are specified to be case-insensitive, and clients and servers are free to mutate the case of single characters (e.g. make it all uppercase, or increase entropy by randomly switch the case of characters) - you may want to read through https://developers.google.com/speed/public-dns/docs/security)

The check for domain name fragments in `to_key` is not sufficient: arbitrary control sequences (e.g. ASCII < 0x20) are left in the strings, as well as whitespaces and '.' characters.  I struggle to understand why there is need for both `to_string` and `to_key` (one which uses '.' as separator, the other '\000'). EDIT: see [RFC2181[(https://tools.ietf.org/html/rfc2181) for some clarifications, it is a bit more tricky (valid names for the internet vs valid data in the caching key-value store).

I suspect the authoritative nameserver code is good as it goes since (zone) input is always trusted (threat model does not include to get something bad here), and user questions are looked up in the `trie` data structure (which uses `to_key`, and thus lowercased strings).

To get closer to the RFC, we should preserve the case (of questions. in answers -- a question: A foo.bAr.com should reply with q: foo.bAr.com answer: foo.bAr.com 300 IN|A 1.2.3.4 -- whereas atm the reply is: q: foo.bAr.com answer: foo.bar.com).

Is there anyone working in this direction (or having resources to do so?), or best left as it is, since it is shipped and deployed code which "works for me"?

EDIT: rephrased some paragraphs

avsm:
This is a good issue to raise, as ocaml-dns is a really old library now -- over 11 years old for some of the original code, and it has been through multiple iterations of parsing techniques as our library ecosystem has grown.

To this:
[...]

PR #143, opened 2017-06-15
Port to jbuilder

### LINKED ISSUE #114: mirage-dns should be dns-mirage

### LINKED ISSUE #135: dns.lwt dependencies in opam have gone crazy
To get dns.lwt, we now need to install these packages in opam:

```
             "dns"  { >= "0.15.3" }
+            "mirage-time" "mirage-stack-lwt" "mirage-kv-lwt" "duration" "mirage-profile"
+            # above are needed for dns.lwt
```

That's a lot to digest.  We should have a `dns-lwt` package instead that captures these dependencies concretely.

avsm:
Rearrange the `opam` packages to eliminate optional dependencies, and have explicit and separate packages for the Lwt, Async and Mirage implementations.  The `opam` and `ocamlfind` layouts now have the same names:
    
- `dns`: the core package
- `dns-lwt`: Lwt implementation
- `dns-lwt-unix`: Lwt Unix, including servers
- `dns-async`: Async implementation (this currently uses Unix)
    
This layout is not compatible with the older releases which had ocamlfind subpackages, so an upper bound will be needed in OPAM for those. However, porting should be relatively straightforward to the new release, and in return your users will not have to deal with specifying a myriad of optional dependencies in OPAM.

Notes:

- This closes #135 
- We could also fix #114 as part of this series -- should I rename `mirage-dns` to `dns-mirage` for consistency?
- This is on the path to fixing #137 as it significantly simplifies the build and modernises it


avsm:
I'm going to try and make this work without any transitionary packages and introduce upper bounds into the repository instead. That'll hopefully work better since it eliminates depopts testing altogether...

Issue #146, opened 2017-10-16
dns-async raises exception "writer fd unexpectedly closed".


1yefuwang1:
After cloning, I type 'make test' on command line, and the following is what I get:

> 2017-10-16 14:19:16.434280+08:00 5.153.225.51
2017-10-16 14:19:16.440170+08:00 Error ("Exception raised to [Monitor.try_with] that already returned.""This error was captured by a default handler in [Async.Log]."(exn(monitor.ml.Error("Writer error from inner_monitor""**writer fd unexpectedly closed** "(writer((id 2)(fd((file_descr 5)(info(replaced(("connected to"8.8.8.8:53)(previously_was(socket((family((family PF_INET)(address_of_sockaddr_exn <fun>)(sexp_of_address <fun>)))(socket_type SOCK_DGRAM)))))))(kind(Socket Active))(supports_nonblock true)(have_set_nonblock true)(state Closed)(watching((read Not_watching)(write Not_watching)))(watching_has_changed false)(num_active_syscalls 0)(close_finished Empty)))(monitor(((name(id 16))(here())(id 16)(has_seen_error false)(is_detached false))((name try_with_or_error)(here())(id 9)(has_seen_error false)(is_detached true))))(inner_monitor(((name(id 17))(here())(id 17)(has_seen_error true)(is_detached true))((name try_with_or_error)(here())(id 9)(has_seen_error false)(is_detached true))))(background_writer_state Stopped_permanently)(background_writer_stopped(Full()))(syscall Per_cycle)(bytes_received 28)(bytes_written 0)(scheduled_bytes 0)(scheduled_back 0)(back 0)(close_state Open)(close_finished Empty)(close_startedEmpty)(num_producers_to_flush_at_close 0)(flush_at_shutdown_elt(<opaque>))(check_buffer_age(((writer <opaque>)(maximum_age2m)(bytes_received_at_now_minus_maximum_age 0)(bytes_received_queue())(times_received_queue())(bytes_seen 0)(too_old Empty))))(consumer_left Empty)(raise_when_consumer_leaves true)(open_flags(Full(Ok(rdwr nonblock))))(line_ending Unix))))("Caught by monitor (id 16)"))))

[...]

PR #147, opened 2017-10-16
close reader and writer properly to avoid raising exception unexpectedly

1yefuwang1:
this commit fixes ISSUE 146


djs55:
Thanks for the PR!

Unfortunately I don't understand the async resolver well enough to review it properly. Does anyone else have any comments?

djs55:
I see the `make test` unit tests reproduces the bug in #146 (and so the unit tests are broken) but this patch fixes them. I'll take a closer look...

djs55:
OK, I agree this patch does a better job of cleaning up than the original code (which never closed the `Reader` or the `Writer`). I'll merge it.

Thanks again!

PR #154, opened 2018-05-29
Correction on parsing function : pattern matching is not correct, total length is not checked

Willy-Tan:
Domain name labels can be either be empty (root), text labels or pointers. The first two bits help to distinguish text labels from pointers : text labels start with two 0 bits (e.g. 0b0_00000001), whereas pointers start with two 1 bits (e.g. 0b0_11000001). That means, any label starting with 10 or 01 is invalid.

However, the parsing function in name.ml did it wrong when checking for pointers. It only checks if one of the two first bits are 1 to assume it's a pointer :
`let v = 0b0_10000001 in (v land 0b0_11000000) != 0` returns `true` even though v isn't a pointer.

A solution would be to replace 
`(v land 0b0_11000000) != 0` 
with 
`(v land 0b0_11000000) == 0b0_11000000`.

Another solution could be `v >= 0b0_11000000`.

### DIFF
--- a/lib/name.ml
+++ b/lib/name.ml
@@ -51,7 +51,7 @@ let to_string = String.concat "."
 (* TODO: this looks wrong for the trailing dot case/we should ensure
    we handle the trailing dot case consistently *)
 let of_string (s:string) : t =
-  Re_str.split (Re_str.regexp "\\.") (String.lowercase_ascii s)
+  Re.Str.split (Re.Str.regexp "\\.") (String.lowercase_ascii s)
 let string_to_domain_name = of_string
 
 let of_ipaddr ip = of_string_list (Ipaddr.to_domain_name ip)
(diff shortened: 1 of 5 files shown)


lib/name.ml
```diff
@@ -67,7 +67,7 @@ let parse_label base buf =
     | 0 ->
         Z base, 1
 
-    | v when ((v land 0b0_11000000) != 0) ->
+    | v when ((v land 0b0_11000000) == 0b0_11000000) ->
```
avsm: I believe this can just be `=` since `==` is to test for physical equality (it shouldnt matter on integers)

Willy-Tan:
I have added some changes to the parsing function :

RFC2181 states that there are only two restrictions on domain name syntax : a label length restriction, which is checked in Dns.Name.parse_label, and a total length restriction which isn't checked in Dns.Name.parse. 

The total length of a domain is the sum of each label length with the separators and the trailing dot included (in fact the restriction is on total byte length but the whole string length is equal to the byte one if we take into account the trailing dot). This length should be less than 255.

I changed Dns.Name.parse to take into account the total length by adding a `size` argument to the folding recursive function. It is incremented by : 
- 1 if the current label is the ending dot (zero octet)
- [label length] + 1 (buffer offset variation when reading a label) if it is a whole label
- [length of pointed labels] + [number of pointed labels] if it is a pointer

I have made the choice to check the size restriction after the recursive function has been called so that the total invalid length can be caught. but it could be more efficient to raise an error within the recursive function so that it stops as soon as the restriction is met.

I also changed somewhere Re_str (apparently deprecated) by Re.Str.

Willy-Tan:
I am putting other modifications here as it is not merged yet. 

Found another error while fuzzing : `Dns.Packet` behaviour for opcodes is not consistent. According to standards, opcode can take 16 values, and only 5 of them are assigned : 
- 0 : `Query` (named `Standard` in ocaml-dns)
- 1 :` IQuery` (deprecated, named `Reverse` in ocaml-dns)
- 2 : `Status` 
- 4 : `Notify`
- 5 : `Update`
- 3 or 6-15 : `Reserved` (for future use)

However, the current implementation only considers `opcode=3` to be `Reserved`. It leads to incorrect behaviours : the parsing function doesn't raise an error when opcode is set to 3 and returns `Reserved`, but raises an error when it tries to parse other reserved opcodes. To solve this, i changed 
`type opcode = Standard | Reverse | Reserved | Notify | Update`
to 
`type opcode = Standard | Reverse | Notify | Update | Reserved of int`

I had to erase the `cenum` macro as `Reserved` is no more constant. For that reason, I rewrote some of the automatically generated functions, and changed `int_to_opcode` to return an `opcode` and not an `opcode option` -- I'm not sure if it should be done but it seemed incorrect with my changes.

I also changed deprecated `Re_str` to `Re.Str` in `Dns.Probe`.

djs55:
The CI is green apart from alpine which failed in another package:
```
#=== ERROR while installing core.v0.11.1 ======================================#
# opam-version         1.2.2
# os                   linux
# command              jbuilder build -p core -j 4
# path                 /home/opam/.opam/4.04.1/build/core.v0.11.1
# compiler             4.04.1
# exit-code            1
# env-file             /home/opam/.opam/4.04.1/build/core.v0.11.1/core-10738-ea44d2.env
# stdout-file          /home/opam/.opam/4.04.1/build/core.v0.11.1/core-10738-ea44d2.out
# stderr-file          /home/opam/.opam/4.04.1/build/core.v0.11.1/core-10738-ea44d2.err
### stderr ###
# unix_time_stubs.c:50:1: warning: control reaches end of non-void function [-Wreturn-type]
# [...]
# unix_stubs.c:336:2: warning: #warning "_POSIX_SYNCHRONIZED_IO undefined or <= 0; aliasing unix_fdatasync to unix_fsync" [-Wcpp]
#  #warning "_POSIX_SYNCHRONIZED_IO undefined or <= 0; aliasing unix_fdatasync to unix_fsync"
#   ^~~~~~~
# unix_stubs.c:1373:2: warning: #warning "_POSIX_PRIORITY_SCHEDULING not present; sched_setscheduler undefined" [-Wcpp]
#  #warning "_POSIX_PRIORITY_SCHEDULING not present; sched_setscheduler undefined"
#   ^~~~~~~
#     ocamlopt src/.core.objs/core__Iobuf.{cmx,o}
# File "src/iobuf.ml", line 1033, characters 4-62:
# Warning 55: Cannot inline: Function information unavailable
```

avsm:
Actually, I'll put this into ocaml-dns.1.1.0. Thanks for the hard work on this @Willy-Tan !

PR #155, opened 2018-10-05
Use Dune

samoht:


samoht:
/ping @djs55 and @avsm (and @emillon :p)

examples/dune
```diff
@@ -0,0 +1,9 @@
+(executables
+  (names     forward server)
+  (libraries dns-lwt-unix))
+
+(alias
+  (name    runtest)
+  (deps    forward.exe server.exe)
+; (action  (run ${<} -e -v))
```
emillon: you can probably convert that to (tests), I think it supports (deps).

lib_test/ounit/dune
```diff
@@ -0,0 +1,9 @@
+(executable
+  (name       test)
+  (libraries  dns ounit pcap-format)
+  (preprocess (pps ppx_cstruct)))
+
+(alias
+  (name    runtest)
+  (deps    test.exe (glob_files *.pcap) (glob_files *.zone))
+  (action  (run ${<})))
```
emillon: In a dune file, that probably tries to run the wrong thing.

samoht: ha yes, good catch

Makefile
```diff
@@ -1,32 +1,10 @@
...
-# until we have https://github.com/ocaml/opam-publish/issues/38
-pkg-%:
-	topkg opam pkg -n $*
-	mkdir -p $(PACKAGES)/$*
-	cp -r _build/$*.* $(PACKAGES)/$*/
-	cd $(PACKAGES) && git add $*
-
-PKGS=$(basename $(wildcard *.opam))
-opam-pkg:
-	$(MAKE) $(PKGS:%=pkg-%)
```
djs55: Out of curiosity, what's the recommended way of creating+publishing packages?

examples/dune
```diff
@@ -0,0 +1,9 @@
+(executables
+  (names     forward server)
+  (libraries dns-lwt-unix))
+
+(alias
+  (name    runtest)
+  (deps    forward.exe server.exe)
+; (action  (run ./forward.exe -e -v))
```
djs55: Is the `;` commenting-out the action?

lib_test/unix/dune
```diff
@@ -0,0 +1,9 @@
+(executable
+  (name      time_server)
+  (libraries dns-lwt-unix))
+
+(alias
+  (name   runtest)
+  (deps   time_server.exe)
+;  (action (run %{deps} --help))
```
djs55: If this is supposed to be commented-out maybe we should add a comment explaining why.

Makefile
samoht: you can use [dune-release](https://github.com/samoht/dune-release). That's not perfect but does relatively well in general :-) Feel free to open issues if that's not the case.

examples/dune
samoht: yes, it was like that before; I guess it's because that test needs the server to run first?

djs55: Ah fair enough, I didn't realise it was already there

PR #156, opened 2018-10-05
Use the new KV.RO API

samoht:
On top of #155


hannesm:
this is good to go. I'd merge and release (should I release only mirage-dns, or all the packages /cc @avsm)!?) unless someone objects.

PR #159, opened 2019-04-04
prepare dns 2.0: merge with udns

avsm:
This replaces the existing implementation of ocaml-dns with two years of work by @hannesm into building a modern DNS stack.  This PR is a subtree merge of the [udns](https://github.com/roburio/udns) repository, with the intention of releasing it as as dns-2.0 suite of packages and deprecating the existing code.

udns is written with modern coding standards and security in mind, so is generally a better and more maintained and correct implementation.  The major deviations from the dns.1.x series of code is:

- [ ] multicast DNS is currently not supported. document this limitation as a future feature.
- [x] `Resolvconf` is missing. Why was it in ocaml-dns in the first place and who uses it? *(wontfix: see https://github.com/mirage/ocaml-dns/pull/159#issuecomment-483038199)*
- [ ] Async adapter doesn't exist. Doesn't need to be in the first version, but we should search for existing users.
- [ ] handling of unknown resource record types (https://tools.ietf.org/html/rfc3597) is _not_ implemented by uDNS, already the 'decode' will error if unknown. There are multiple reasons for this behaviour: in the "simple" resolver case, we do not want to forward bad data, and comparison of RRs is far simpler with this model.
- [ ] Are there any missing RR types used in the wild that aren't currently implemented?
- [ ] there's no wildcard DNS entries support for the server. it's rather unlikely this will ever be supported, so the limitation needs to be documented.
- [x] remove the .merlin files

Before a merge here, we also need to:

- [x] rename `Udns` modules to `Dns`
- [x] rename `udns-*` packages to `dns-*`
- [ ] post this plan to discuss.ocaml.org as a PSA
- [x] double check my merge to ensure the udns history and current snapshot is correct. I've verified this with a checkout and `diff -uNr` of the two repositories.
- [x] update the README to reflect the changes and perhaps a migration guide from dns 1.x to 2.x *(see e430c6b1f8c1e8b3bee9498b57263cf0dc39cbd8)*
- [x] verify that major consumers of dns such as Mirage and Vpnkit (@djs55) can work with 2.0 *(they dont work with vpnkit directly but djs55 is working on a port)*

Opening this PR up for any discussion and objections to this approach.  It has previously been floated as an idea with @hannesm and @djs55, and we came up with the above list.  More feedback welcome!


hannesm:
thanks @avsm. the list of items above show only the downsides :(, while there are as well some upsides of uDNS:
- its data (rrset) is defined in a single GADT https://roburio.github.io/udns/doc/udns/Udns/Rr_map/index.html#type-k
- it supports (in contrast to ocaml-dns): notify, dynamic update, zone transfer, tsig (hmac authentication), edns
- there's no use of `mutable`
- it uses the independent [domain_name](https://github.com/hannesm/domain-name) library which uses string array (instead of string lists) for storing domain names, improving the performance (not sure where my benchmarks have gone :/)
- there's integration with [let's encrypt](https://github.com/mmaker/ocaml-letsencrypt) for provisioning valid X.509 certificates
- see as well this (slightly outdated) blog entry https://hannes.nqsb.io/Posts/DNS#What-is-DNS
- please note that @cfcs worked on this as well, mainly designing and implementing the [client API](https://roburio.github.io/udns/doc/udns-client/Udns_client_flow/Make/index.html) which has unix, lwt, and mirage implementations
- more strict hostname validation (see https://github.com/hannesm/domain-name/issues/4, it should be revised a bit)
- I'm running this code as authoritative NS for several domains since 1.5 years :)

in respect to "using it with vpnkit/docker" (which does a caching stub resolver IIRC), here i suspect that there's need for handling unknown RR types (as RFC 3597 says), otherwise there'll be users who attempt the brand new YYY record type which is not supported (yet) and complain. I'm not entirely sure how to integrate that in the best way.

avsm:
I had a word with @djs55 who agrees this good to merge and then work on in master.  I'd like some explicit approvals before doing so, ideally from @hannesm and @djs55 at least. 

I've merged from latest udns and the only diverging changeset should be the CHANGES and README if you diff the trees.  Since history is preserved, my merge was a standard `git pull` from the roburio/udns git remote.  Once merged into master, I can work on renaming `Udns` to `Dns`, but this should perhaps be done in the `udns` remote if @hannesm prefers to keep working there for a while.

avsm:
I investigate the `Resolvconf` parser in the 1.x series, and it is quite ancient (556f695e9eac6ffc3d0b981cd8ed54ff83eb0e9d from the earliest implementation of [Melange](http://anil.recoil.org/papers/2007-eurosys-melange.pdf)). It can easily be moved out into a separate library as it isn't particularly tied to the DNS wire protocol, so I'll mark that in the main PR body as "wontfix" in 2.x

djs55:
Yes indeed, I'm quite keen to see this merged and am happy to work on it in master.

hannesm:
I spend some more time on uDNS - mainly the resolver and the core Packet.t data structure. [This](https://github.com/roburio/udns/pull/23) is pending a review (likely tomorrow after sleeping) to convince myself that there's not much regression introduced.

Apart from that, I'd like to spend this week a bit on documentation (which usually bring me back to code changes ;). I'd be fine with moving the repo over here next weekend (WDYT, @cfcs?). And yes, I can do the renamings to drop the "u" again. Coincidentally, I think I found a solution for unknown record types...

cfcs:
@hannesm I don't think I have a commit bit for this repo :-(

cfcs:
From the client API (or implementation, rather) perspective, I have some things I would like to discuss or solve before launching it into production:
1) The issue of re-using connections instead of making new ones for each query; I also don't think I am closing them at the moment.
2) Should the user be able to mandate DNSSEC for some queries? It seems to me that there a bunch of cases where you want DNSSEC to be checked before trusting the information, like when asking for TLSA records, and if the upstream is not configured correctly for some reason I would prefer a hard failure. Asking now because I would prefer not having to unnecessarily breaking the interface once people have started using it. I imagine something like a named argument to `Udns_client_flow.getaddrinfo -> ?(dnssec = false)` for making individual queries, and a parameter in the NS configuration that would result in *only* returning answers authenticated using DNSSEC.
3) Timeouts & retries

If we want Unix-specific handling of things:

4) Potential handling of `/etc/hosts` (Unix) and `HOSTS` (Windows), seems simple enough?

5) If compiling a mirage unikernel with `-t unix` I want it to behave as it would on other targets; if we implement `/etc/hosts` or `/etc/resolv.conf` I **do not** want it to go off and rely on things in `/etc`. Not sure how to best deal cleanly with that from inside the library?

6) The issue of `resolv.conf` handling as @avsm pointed out (and like we discussed earlier today) for Unix-based hosts. Basically `resolv.conf` seems to be a mess, and things are slightly different on each of the platforms I checked ([debian](https://dyn.manpages.debian.org/stretch/manpages/resolv.conf.5.en.html), [freebsd](https://www.freebsd.org/cgi/man.cgi?query=resolv.conf&sektion=5&apropos=0&manpath=FreeBSD+12.0-RELEASE+and+Ports), [openbsd](https://man.openbsd.org/resolv.conf.5)), and kind of tied to `libresolv` internals. While the are some good ideas here, I am slightl wary as to the value of trying to be bug-for-bug compatible with the commonly accepted subset.
  The options **I** think would make sense to have (in a hypothetical more ideal config format) are:
    - timeouts
    - number of retries
    - upstream nameserver addresses
      - whether to use tcp/udp/tls
        - optional TLS trust root (key pin/CA cert)
      - a way to prevent sending hostname queries (single-label queries) to this upstream
    - whether to accept local IPs for non-hostname (multi-label) names (e.g. do we return `127.0.0.1` for `curl.me.example.com` or not)

As for `resolv.conf` we could probably get by just parsing `nameserver` and the *"use tcp"* directives and ignoring everything else, and then the user could run and configure a local resolver to their wishes. This has the benefit of behaving at least somewhat like what sysadmins would expect from other applications, which is something to consider IMHO.

The commonly supported subset of syntax/entries seem to be:
  - `"sortlist" {CIDR 1} {CIDR 2} {CIDR .. 10}`: Unclear if this can appear multiple times, but basically this seems to define an ordering of IP addresses for `gethostbyname()`, although it wasn't clear to me from the documentation how this works. Do the listed CIDR ranges sort first or last, relatively and compared to unlisted ranges? This seems useful, but I've never seen anyone use this.
  - `"nameserver" {IP}`: Can appear as multiple entries 1, 2, or 3 times in `resolv.conf`. Specify upstream nameserver in IPv4 (quad dot notation) or IPv6 (hex-and-colon notation) address format. The manpages suggest that handling is somewhat uniform: If no nameserver is given, "the nameserver on the local machine" (whatever that means?? 127.0.0.1?) is used. They all seem to choose a NS at random(? on debian `options rotate` suggests that it chooses them in the order they're listed in the file) and wait for a timeout before trying the next. I would suggest dropping the "nameserver on the local machine" thing, not clear to me how to determine the address of that, and I would go with a hardcoded upstream NS instead in the case where there is no configured `nameserver`. 
  - `"options"`
    - "use TCP instead of UDP": `"usevc"`(freebsd) `"tcp"` (openbsd) `"use-vc"` (debian)
    - `"ndots"`: never seen this used, but they all have this. something to do with `search list`.
  - `"domain"`: I've never used this, so not sure how this works, but it's some option for appending one or more (?) labels to names before resolving, under some circumstances. They seem to agree, and that if not present it should default to the excess labels from the result of `gethostname(2)`. Seems useful if we can figure out the semantics, ie does this only kick in if I try to resolve `hello`, or will `hello.example.com` somehow result in a query for `hello.example.com.mycooldomain` if the `hello.example.com` query fails?
  - `"search" {domain 1} {domain 2} {domain .. 5}`: seems to extend the `domain` entry(/*ies*?) with up to five extra entries, modulo some magic with `ndots`. Seems useless to me.

djs55:
Regarding resolving DNS names in Unix apps e.g. with `mirage -t unix` are there any platform-specific libraries we could use rather than rewrite the logic ourselves here? For example on Mac we could use the `DNSServiceRef` API (bound here: https://github.com/mirage/ocaml-osx-dnssd ) to perform lookups which will respect Mac-specific stuff like [/etc/resolver](https://apple.stackexchange.com/questions/74639/do-etc-resolver-files-work-in-mountain-lion-for-dns-resolution). On windows the equivalent API seems to be [DnsQuery](https://docs.microsoft.com/en-us/windows/desktop/api/windns/nf-windns-dnsquery_a). In the moby/vpnkit project I tried to reimplement the resolution policy on Mac and Windows but I gave up in the end and bound these platform APIs instead. Perhaps a `mirage -t unix` unikernel could use a system-specific resolver library at configure-time (if available) and fall back to something really simple and generic if there's nothing available.

cfcs:
@djs55 I think for `mirage  -t  *` we should strive to achieve uniform platform-independent behaviour, so in that case having consistent logic would be preferable IMHO.

I brought `resolv.conf` up in relation to the "real" (without Mirage) `unix` case, where something like that could be used, but it seems like that one does more than parse the local policies though - it actually performs (blocking) requests?

It feels like that project has a different scope/aim than this cross-platform library, and I'm a little bit uncomfortable introducing dependencies on C code for commercial platforms that few people have access to, especially when the library being bound to has had several remotely exploitable flaws in the past, ie CVE-2015-7015, and we here have a pure-OCaml implementation that will hopefully protect us against things like that.

avsm:
@cfcs - you now have access to the repo. Welcome to mirage! :-)

With respect to this library, I think it should stay focussed on the DNS _protocol_ resolution -- so iterative or recursive network requests.  The issue of client stub resolver integration with existing systems is complex and non-portable and better off in a separate library.

So, to split up the questions above from @cfcs, I think we should:
- from @cfcs' list, solve 1 2 3 in this library since it affects DNS-on-the-wire resolution
- defer Unix parsing of resolv.conf to a separate library `ocaml-resolv-conf` which can deal with all the OS specific semantics: see https://github.com/tailhook/resolv-conf for something we could use as a guide for the full set of OS specific pieces
- just keep the direct C bindings to (e.g.) glibc and osx in a separate `dns-stub-resolver` library which introduces the C dependencies and exploit surface from that usually bugridden code.

Since it seems like we have consensus on moving here, I'll also merge this later on tonight to the `master` branch and then post a PSA to discuss.ocaml.org to see if we can get a volunteer for the async backend.  Then we can close out a number of the older issues and have the points raised here on the issue tracker (such as the udns rename and so on).

samoht:
I like this proposal 👍

hannesm:
as status update, I revised some bits and pieces of udns again, and today merged also the `Udns -> Dns` and `udns* -> dns*` package rename... of course there are still some bugs hidden in udns ;)

avsm:
It is done! Now converting the remaining TODOs into tracking issues

avsm:
@hannes, just to confirm -- you did add support for unknown RRs with the refactoring of `rr` in 4626747074b5eca2f5a4ce60553c47346ad54250 right?  I'll mark that as complete in the above list for now.

PR #164, opened 2019-06-28
Updates (README, domain-name, ..)

hannesm:
this are cherry-picks from the udns repository. it is slightly confusing now to have both repositories with different git history, and some issues tracked in either of them :/ (I closed several issues in this repository)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Build Status](https://travis-ci.org/mirage/ocaml-dns.svg?branch=master)](https://travis-ci.org/roburio/udns)
 
-(c) 2017,2018 Hannes Mehnert (robur.io, Center for the Cultivation of Technology)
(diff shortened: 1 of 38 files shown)


cfcs:
I say LGTM, but I'm a bit worried about the two repositories drifting apart, increasing the maintenance burden.

Perhaps the easiest solution would be to just archive the other mirror and move development here?

hannesm:
@cfcs yes, good idea.

hannesm:
could you approve, so that I'm able to merge this? ;) (in this repository, approval is required!)

cfcs:
Sorry, yes, I hadn't noticed :)

hannesm:
I archived the roburio/udns.git repository and opened various issues in this repository, some with milestone 2.0.0. if I missed something, please open new issues (and assign milestone 2.0.0 if we should have this before an initial public release). would be great to get an initial public release in july 2019!

PR #173, opened 2019-07-15
sync CHANGES with the 1.1 branch

avsm:


hannesm:
not sure whether there is value, since 1.x only lives in their branch... may be worth to note that in changes?

avsm:
I think it's worthwhile having a single CHANGES file that has all the version information. What's the point in branching it?

PR #176, opened 2019-07-19
fixes #110

### LINKED ISSUE #110: gethostbyname with socket options
Hello, is it a simple way to use `Dns_resolver_unix.gethostbyname` but with some modifications on the created socket? For example I want to bind the created socket on a specific address (like -b option in dig), but it seems that I need to rewrote the code written in `Dns_resolver_unix` and use `Dns_resolver` module for that.

Is there a better option to do this?

cfcs:
1) FWIW this is one possible way to solve #110 
    - EDIT: Well, besides `Unix.setsockopt` not allowing you to bind a socket to an interface. You can bind it to an address using `Unix.bind`, but that doesn't guarantee anything about the interface unless you're using raw sockets, which `Unix.sockaddr` does not expose.
    - EDIT 2: Perhaps you can `Unix.bind` to a specific device with an IPv6 address with zone id `[::%eth0]`, this is maybe the most portable you can get. Not sure if the stdlib supports that.

2) Another approach could be to instead take an already connected socket (which the application can then setup however it wants) - (EDIT: With this kind of interface you are free to pursue platform-specific ways to bind sockets to interface, e.g. `SO_BINDTODEVICE` on Linux):
```ocaml
  type io_addr =
    | `Fd of Unix.file_descr
    | `Address of Unix.inet_addr * int
```

3) That however becomes problematic if we ever want to do automatic reconnections. In that case one could conceivably have:
```ocaml
  type io_addr =
    | `Connect_function of (unit -> (Unix.file_descr, [> `Msg of string]) io)
    | `Address of Unix.inet_addr * int
```

I don't have a use-case for socket options, so I don't have an opinion.
3) seems to me like the most portable interface to allow the same for other backends since that could work for `Lwt` / `Mirage` too without too many modifications.

ping @jerome-arzel and @hannesm

### DIFF
--- a/unix/client/dns_client_unix.ml
+++ b/unix/client/dns_client_unix.ml
@@ -9,14 +9,19 @@ module Uflow : Dns_client_flow.S
    and type stack = unit
    and type (+'a,+'b) io = ('a,[> `Msg of string]as 'b) result
 = struct
-  type io_addr = Unix.inet_addr * int
+  type io_addr =
+    [ `Bool of Unix.socket_bool_option * bool
+    | `Int of Unix.socket_int_option * int
+    | `Intopt of Unix.socket_optint_option * int option
+    | `Float of Unix.socket_float_option * float
+    ] list * Unix.inet_addr * int
   type ns_addr = [`TCP | `UDP] * io_addr
   type stack = unit
   type flow = Unix.file_descr
   type t = { nameserver : ns_addr }
   type (+'a,+'b) io = ('a,'b) result constraint 'b = [> `Msg of string]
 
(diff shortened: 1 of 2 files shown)


hannesm:
I think we settled that to support this, someone could implement their custom Dns_client_flow.S, and if this is valuable the implementation could be upstreamed. the focus for the client API is for now slightly different, namely:
- not leak file descriptors / connections (by providing a close)
- re-using an existing connection instead of one connection per request (to amortise the handshake delay/costs)
- support DNS-over-TLS

if @cfcs agrees with the plan above, we should close this PR.

PR #177, opened 2019-07-23
provide dynamic update of HMAC / TSIG keys

hannesm:
this has been on some branch for a while, the primary server now contains a function to update the keys. Upon a key update, various janitoring is done (potentially notifications are generated (when new keys are added), other notifications are dropped (when keys are removed)).

This functionality is not exposed via `nsupdate` (which was the case in an earlier no longer existing incarnation of uDNS -- with various `_key-management` keys -- a bad idea to transmit secrets in the clear via the wire), but via API. In deployments I use this to dynamically update the keys from a zone file stored in a git remote (to which someone pushes, and the unikernel pulls after getting notified).

two other commits in this PR:
- do not _require_ a DNS key in ocertify -- there are valid uses that only download the (already existing) TLSA record
- instead of calling `Domain_name.to_array` and manually walking through the label array, use the functions of domain-name 0.3.0 (esp. `find_label`)

### DIFF
--- a/app/ocertify.ml
+++ b/app/ocertify.ml
@@ -99,32 +99,29 @@ let jump _ server_ip port hostname dns_key_opt csr key seed bits cert force =
      | Error (`Msg m) -> Error (`Msg m)
      | Error ((`Decode _ | `Bad_reply _ | `Unexpected_reply _) as e) ->
        Error (`Msg (Fmt.strf "error %a while parsing TLSA reply" Dns_certify.pp_q_err e)))
-  >>= fun send_update ->
-  if send_update then
+  >>= function
+  | false -> Ok ()
+  | true ->
     match dns_key_opt with
     | None -> Error (`Msg "no dnskey provided, but required for uploading CSR")
     | Some (keyname, zone, dnskey) ->
       nsupdate_csr sock hostname keyname zone dnskey req >>= fun () ->
       let rec request retries =
-        if retries = 0 then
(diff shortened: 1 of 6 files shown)


hannesm:
CI is green, @cfcs would be nice to get this merged (towards a release...) :)

app/ocertify.ml
```diff
@@ -101,25 +101,28 @@ let jump _ server_ip port (keyname, zone, dnskey) hostname csr key seed bits cer
...
+    | Some (keyname, zone, dnskey) ->
+      nsupdate_csr sock hostname keyname zone dnskey req >>= fun () ->
+      let rec request retries =
+        if retries = 0 then
+          Error (`Msg "failed to request certificate")
+        else
+          match query_certificate sock public_key hostname with
+          | Error `No_tlsa ->
+            Logs.warn (fun m -> m "still no tlsa, sleeping two more seconds");
+            Unix.sleep 2;
```
cfcs: If I'm reading this correct the sleep will happen also in the case where `retries = 1`, then it will recurse with `retries = 0` which means it'll hit the `failed to request certificate` path, ie waiting for 2 seconds without anything productive to happen after that time, even if `query_certificate` now would return OK.

What do you think about something like this at the top of the pattern match instead?
```ocaml
| Error _ when retries = 0 -> Error (`Msg "failed to request certificate")
```

hannesm: i hope 5be954d satisfies you :)

PR #181, opened 2019-08-04
Dns_client_flow: require "close : flow -> (unit, 'err) io" to close a flow

hannesm:
Dns_*client*: provide close, and close on error to avoid file descriptor leaks

atm, the definition of `type (+'ok,+'err) io` and `map` make it impossible to handle potential `recv`/`send` errors within Dns_client_flow (i.e. tear down/close the socket). if it'd be `type +a io` (and `map` being common Lwt.bind, i.e. the result type being external), this would be easier to do. WDYT?


hannesm:
(the alternative described above is implemented in #182, which I prefer over this PR (since the close is at more obvious locations [unfortunately in case `connect` fails, we need to close in lwt+unix manually (since we don't return that opened `socket` to the caller)]))

hannesm:
superseeded by #182

PR #182, opened 2019-08-04
Dns_client: revise "io" type (be polymorphic only in 'a, not in ('a, 'b))

hannesm:
Dns_client_flow: require "close : flow -> unit io" to close a flow
Dns_*client*: provide close, and close on error to avoid file descriptor leaks

this is meant as alternative to #181

### DIFF
--- a/unix/client/dns_client_unix.ml
+++ b/unix/client/dns_client_unix.ml
@@ -7,25 +7,26 @@ module Uflow : Dns_client_flow.S
   with type flow = Unix.file_descr
    and type io_addr = Unix.inet_addr * int
    and type stack = unit
-   and type (+'a,+'b) io = ('a,[> `Msg of string]as 'b) result
+   and type +'a io = 'a
 = struct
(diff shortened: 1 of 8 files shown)


unix/client/dns_client_unix.ml
```diff
@@ -35,23 +36,34 @@ module Uflow : Dns_client_flow.S
     end >>= fun proto_number ->
     let socket = Unix.socket PF_INET SOCK_STREAM proto_number in
```
cfcs: - `Unix.socket` can also fail, we should probably guard that too. Technically speaking I think the same goes for `getprotobyname` which I guess can fail if e.g. the application is running from within a chroot or sandbox that doesn't have access to `/etc/protocols`.
  - Same for the equivalent `Lwt` wrappers in `lwt/client/dns_client_lwt.ml`

hannesm: well spotted, I implemented these guards in 14c6859

PR #183, opened 2019-08-04
dns_server: more comments, less long lines

hannesm:


server/dns_server.ml
```diff
@@ -40,11 +41,12 @@ module Authentication = struct
...
-    try
-      let rec go idx = if Array.get arr idx = transfer then idx else go (succ idx) in
-      let zone_idx = go 0 in
+    let is_transfer = Domain_name.equal_label (operation_to_string `Transfer) in
+    match Domain_name.find_label ~rev:true name is_transfer with
+    | None -> None
+    | Some idx ->
+      let arr = Domain_name.to_array name in
+      let zone_idx = Array.length arr - succ idx in
       let zone = Domain_name.of_array (Array.sub arr 0 zone_idx) in
```
cfcs: now that we have `drop_label` I think this can be solved with something like (or `~rev:true` if we want the other side?)
```ocaml
let zone = Domain_name.drop_label_exn ~amount:(succ idx) d in
```

server/dns_server.ml
```diff
@@ -461,13 +477,13 @@ module Notification = struct
...
       end else begin
-        Log.warn (fun m -> m "replacing key zone %a oldkey %a and IP %a, new key %a"
+        Log.warn (fun m -> m "replacing key zone %a oldkey %a IP %a, new key %a"
                      Domain_name.pp zone Domain_name.pp k Ipaddr.V4.pp ip
                      Domain_name.pp key);
         cs'
       end
     | Some None ->
-      Log.info (fun m -> m "adding zone %a with key %a and IP %a (previously: no key)"
+      Log.info (fun m -> m "adding zone %a with key %a and IP %a (was no key)"
```
cfcs: Hmm, I think the old was more descriptive?

server/dns_server.ml
hannesm: yes, but in the following code there's still some functionality missing from domain-name, namely "give me a subdomain starting at index y and length l" -- that's why i got lazy and didn't bother to convert this fragment to not use Domain_name.to_array/of_array yet...

server/dns_server.ml
hannesm: it may have been that it did not fit on 80 characters, but i'll revise this as you wish.

Issue #165, opened 2019-06-30
 Client dns packet id uses Random.int


hannesm:
from https://github.com/roburio/udns/issues/9

@cfcs reports:
I took a shortcut here: https://github.com/mirage/ocaml-dns/blob/master/client/dns_client.ml#L20

Leaving this issue here so we don't forget about it.

EDIT: Also it's off by one (should be 0x10000):

```
Random.int bound returns a random integer between 0 (inclusive)
and bound (exclusive). bound must be greater than 0 and less than 2^30.
```

@hannesm further reports:
Server code (and resolver) should also unify the use of random, maybe sth in dns.mli should be around that takes a rng and produces a proper id.

hannesm:
this has been resolved by #184 #185

PR #185, opened 2019-08-07
client: use Randomconv.int16, provide rng via create

hannesm:
this is a solution for #165 -- another could be to fully functorize over Mirage_random.C (but that'd bring us a Mirage_random dependency in the pure client, which I'd avoid) -- let me know what you think about this! :)

### DIFF
--- a/client/dns_client.mli
+++ b/client/dns_client.mli
@@ -18,18 +18,12 @@ type 'key query_state constraint 'key = 'a Dns.Rr_map.key
 *)
 
 val make_query :
-  Dns.proto -> 'a Domain_name.t ->
+  (int -> Cstruct.t) -> Dns.proto -> 'a Domain_name.t ->
   'query_type Dns.Rr_map.key ->
   Cstruct.t * 'query_type Dns.Rr_map.key query_state
-(** [make_query protocol name query_type] is [query, query_state]
+(** [make_query rng protocol name query_type] is [query, query_state]
     where [query] is the serialized DNS query to send to the name server,
-    and [query_state] is the information required to validate the response.
-
-    NB: When querying for [TLSA] records, it is important to use the optional
-    [~hostname:false] parameter with the conversion functions within {!Domain_name}
-    when constructing the {!Domain_name.t} for the search, since these contain
-    labels prefixed with underscores.
-*)
+    and [query_state] is the information required to validate the response. *)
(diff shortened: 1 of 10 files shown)


client/dns_client.mli
```diff
@@ -18,18 +18,12 @@ type 'key query_state constraint 'key = 'a Dns.Rr_map.key
...
   Cstruct.t * 'query_type Dns.Rr_map.key query_state
-(** [make_query protocol name query_type] is [query, query_state]
+(** [make_query rng protocol name query_type] is [query, query_state]
     where [query] is the serialized DNS query to send to the name server,
-    and [query_state] is the information required to validate the response.
-
-    NB: When querying for [TLSA] records, it is important to use the optional
-    [~hostname:false] parameter with the conversion functions within {!Domain_name}
-    when constructing the {!Domain_name.t} for the search, since these contain
-    labels prefixed with underscores.
```
hannesm: this comment was for domain-name <= 0.2.0, since then it is no longer true (`Domain_name.of_string` now results in a `` [ `raw ] Domain_name.t`` :))

client/dns_client_flow.ml
```diff
@@ -1,3 +1,11 @@
+
+let stdlib_random n =
+  let b = Cstruct.create n in
+  for i = 0 to pred n do
+    Cstruct.set_uint8 b i (Random.int 256)
+  done;
+  b
```
hannesm: could as well be required as a dependency from mirage-stdlib-random (but feels fine to have here imho)

hannesm:
NB: for the current use case, a function `unit -> int` (where the latter int should be something between 0x0000 and 0xFFFF) would be sufficient, but in the future we'll likely want label case randomisation which requires some random bits, thus I'd go with the `int -> Cstruct.t` we have specified and used elsewhere in mirage.. (well, a rng could have a nicer type, but that's sth to think about and fix elsewhere)

client/dns_client_flow.mli
```diff
@@ -4,6 +4,8 @@
          better solution presents itself.
 *)
 
+val stdlib_random : int -> Cstruct.t
```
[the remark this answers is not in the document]
hannesm: I did, and pointed to alternatives -- i.e. Nocrypto.Rng.generate

PR #191, opened 2019-09-26
Dns_client: Add tests, and a cache, rename Uflow into Transport, introduce Dns_client.Pure

olleolleolle:
TL;DR: Add **cache** to the DNS client.

## Details

- Add tests
- rename `Dns_client` into `Dns_client.Pure` to free up the name "Dns_client"
- rename `Dns_client_flow` into `Dns_client` so that users only interact with the obviously-named thing
- rename `Uflow` into `Transport`, arguably easier to understand
- add `Dns_cache` and start using it in the centrally-placed `getaddrinfo` function
- add dependency `lru` 0.3.0, or later

_Authors_: @linse @hannesm @olleolleolle and @cfcs


client/dns_client.ml
```diff
@@ -1,87 +1,212 @@
...
+  type +'a io
+  type io_addr
+  type ns_addr = ([`TCP | `UDP]) * io_addr
+  type stack
+  type t
+
+  val create : ?rng:(int -> Cstruct.t) -> ?nameserver:ns_addr -> stack -> t
+
+  val nameserver : t -> ns_addr
+  val rng : t -> (int -> Cstruct.t)
```
cfcs: TODO: maybe store the type `t` as well

client/dns_client.ml
```diff
@@ -1,87 +1,212 @@
...
+module Make = functor (Transport:S) ->
+struct
+
+  type t = {
+    mutable cache : Dns_cache.t ;
+    clock : unit -> int64 ;
+    transport : Transport.t ;
+  }
+
+  let create ?(size=1000) ?rng ?nameserver ~clock stack =
```
cfcs: note: maybe `16` instead of `1000` for the default size

client/dns_client.ml
```diff
@@ -1,87 +1,212 @@
...
+
+  let getaddrinfo (type requested) t ?nameserver (query_type:requested Dns.Rr_map.key) name
+    : (requested, [> `Msg of string]) result Transport.io =
+    let ts = t.clock () in
+    let domain_name = (Domain_name.raw name) in
+
+    match Dns_cache.get t.cache ts domain_name query_type with
+    | Ok (`Entry B (query_type', value), cache) ->
+      t.cache <- cache;
+      assert (query_type' = Obj.magic query_type);
```
cfcs: todo: less obj magic, alternatively guard with `Obj.is_int` on both.

client/dns_client.ml
```diff
@@ -1,87 +1,212 @@
...
+      Transport.lift @@ Ok(Obj.magic value)
+    | Ok _
+    | Error _ ->
```
cfcs: note: cached lookup failures should be handled here (returned to client) instead of making a request
also note: we should cache lookup failures for that to be useful

client/dns_client.ml
```diff
@@ -1,87 +1,212 @@
...
+      let proto, _ = match nameserver with None -> Transport.nameserver t.transport | Some x -> x in
+      let tx, state =
+        Pure.make_query (Transport.rng t.transport)
+          (match proto with `UDP -> `Udp | `TCP -> `Tcp) name query_type
```
cfcs: todo: normalize this stuff wtf

client/dns_client.ml
```diff
@@ -1,87 +1,212 @@
...
+      in
+      Transport.connect ?nameserver t.transport >>| fun socket ->
+      Logs.debug (fun m -> m "Connected to NS.");
```
cfcs: note: we should have our own logs provider so people can filter this stuff out

client/dns_client.ml
```diff
@@ -1,87 +1,212 @@
...
+      (Transport.send socket tx >>| fun () ->
+      Logs.debug (fun m -> m "Receiving from NS");
+      let rec recv_loop acc =
+        Transport.recv socket >>| fun recv_buffer ->
+        Logs.debug (fun m -> m "Read @[<v>%d bytes@]"
+                        (Cstruct.len recv_buffer)) ;
+        let buf = Cstruct.append acc recv_buffer in
```
cfcs: todo: don't do that if `acc` is empty? (depending on whether we care abou tkeeping the reference)

client/dns_client.ml
```diff
@@ -1,87 +1,212 @@
...
+        match Pure.parse_response state buf with
+        | `Ok x ->
+          let entry = `Entry (Rr_map.B (query_type, x)) in
+          let rank = Dns_cache.NonAuthoritativeAnswer in
+          t.cache <- Dns_cache.set t.cache ts domain_name query_type rank entry;
```
cfcs: note: **should** get a fresh value for `ts` timestamp since we have done i/o

client/dns_client.ml
```diff
@@ -1,87 +1,212 @@
...
+          Transport.lift (Ok x)
+        | `Msg xxx -> Transport.lift (Error (`Msg( "err: " ^ xxx)))
```
cfcs:
```suggestion
        | `Msg xxx -> Transport.lift (Error (`Msg xxx))
```
also: this is where we want to cache some of failures (negative responses) ((not the i/o breakdowns or protocol violations))

client/dns_client.mli
```diff
@@ -1,34 +1,148 @@
...
+      and [query_state] is the information required to validate the response. *)
+
+  val parse_response : 'query_type Dns.Rr_map.key query_state -> Cstruct.t ->
+    [ `Msg of string | `Ok of 'query_type | `Partial ]
+  (** [parse_response query_state response] is the information contained in
+      [response] parsed using [query_state] when the query was successful, or
+      an [`Msg message] if the [response] did not match the [query_state]
+      (or if the query failed).
+
+      In a TCP usage context the [`Partial] means there are more packets.
```
cfcs:
```suggestion
      In a TCP usage context the [`Partial] means there are more bytes to be read in order to parse correctly. This can happen due to short reads or if the server (or something along the route) chunks its responses into multiple individual packets. In that case you should concatenate `response` and the next received data and call this function again.
```

test/cache.ml
```diff
@@ -0,0 +1,111 @@
+open Dns
+
+let empty = Dns_cache.empty 100
```
cfcs: note: would be nice to run all the tests for LRU/caches of different sizes.

test/client.ml
```diff
@@ -0,0 +1,312 @@
...
+       ae 00 06 03 6e 73 32 c0  39 c0 35 00 01 00 01 00
+       02 40 8a 00 04 17 15 f2  58 c0 51 00 01 00 01 00
+       02 40 8a 00 04 17 15 f3  77" in
+
+    (* This `rng` generates zeros, used for the query ID above *)
+    let rng = fun x -> Cstruct.create(x) in
+    let name:'a Domain_name.t = Domain_name.of_string_exn "foo.com" in
+    let _actual, state = Dns_client.Pure.make_query rng `Tcp name Dns.Rr_map.A in
+    match Dns_client.Pure.parse_response state ipv4_buf with
+    | `Ok _ -> () (* TODO: Alcotest TESTABLE for this return value *)
```
cfcs: note: that would be nice

test/client.ml
```diff
@@ -0,0 +1,312 @@
...
+       ae 00 06 03 6e 73 32 c0  39 c0 35 00 01 00 01 00
+       02 40 8a 00 04 17 15 f2  58 c0 51 00 01 00 01 00
+       02 40 8a 00 04 17 15 f3  77" in
+
+    (* This `rng` generates zeros, used for the query ID above *)
+    let rng = fun x -> Cstruct.create(x) in
+    let name:'a Domain_name.t = Domain_name.of_string_exn "foo.com" in
+    let _actual, state = Dns_client.Pure.make_query rng `Tcp name Dns.Rr_map.A in
+    match Dns_client.Pure.parse_response state ipv4_buf with
+    | `Ok _ -> failwith "should have rejected mismatched input" (* TODO: Alcotest TESTABLE for this return value *)
```
cfcs: note: write testable

test/client.ml
```diff
@@ -0,0 +1,312 @@
...
+  let tests = [
+      "unpacks some kind of response", `Quick, unpacks_response;
+      "fails to unpack response with mismatching query ID", `Quick, fails_to_unpack_mismatched;
+  ]
+end
+
+(* {!Transport} provides the implementation of the underlying flow
+   that is in turn used by {!Dns_client.Make} to provide the
+   blocking Unix convenience module:
+*)
```
cfcs: note: I think here we don't do that, instead we implement a custom implementation to mock answers. old comment?

test/client.ml
```diff
@@ -0,0 +1,312 @@
...
+    | None -> Ok default_debug_info
+    | Some(_, mock_responses) -> Ok mock_responses
+
+  let send _ _ =
+    Ok ()
+
+  let recv (mock_responses:flow) =
+    (* let cool = (Cstruct.of_string "hello") in
+    Fmt.epr "dnsclient unix recv: %a\n%!"
+            Cstruct.hexdump_pp cool; *)
```
cfcs: note: clean up cool comments

unix/client/dune
```diff
@@ -10,4 +10,4 @@
   (modules      ohost)
   (package      dns-client)
   (public_name  dns-client.unix)
-  (libraries    fmt dns-client.unix))
+  (libraries    fmt dns-client.unix mirage-clock-unix))
```
cfcs: note: use `mtime`

cfcs:
CI fails due to findlib package vs opam package for `mirage-clock-unix`, but we'll be using `mtime` anyway.

test/client.ml
olleolleolle: Yes an old comment.

PR #193, opened 2019-10-10
adapt to x509 0.8.0 API changes

hannesm:


hannesm:
if this could be approved, the CI should be fine in a bit (and x509 0.8.1 got merged to opam-repository)

PR #195, opened 2019-10-13
Dns_cache: use mutable LRU instead of functional one

hannesm:
as suggested in https://github.com/mirage/mirage-nat/pull/28#issuecomment-482619504
also, reduced complexity (O(log n) -> O(1)) https://pqwy.github.io/lru/doc/lru/Lru/index.html

### DIFF
--- a/cache/dns_cache.ml
+++ b/cache/dns_cache.ml
@@ -91,10 +91,11 @@ end
 
 module Key = struct
   type t = [ `raw ] Domain_name.t
-  let compare = Domain_name.compare
+  let equal a b = Domain_name.equal a b
+  let hash r v = Hashtbl.seeded_hash r v
 end
 
-module LRU = Lru.F.Make(Key)(Entry)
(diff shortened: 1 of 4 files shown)


hannesm:
`get` does a promotion in the LRU (so that the least-recently-used is updated)

hannesm:
@cfcs to be more precise: the caller of `get` used to assign the returned cache https://github.com/mirage/ocaml-dns/pull/195/files#diff-56b338e49668122fb96f9dd700470759L185-L187 (and `get` itself does the promotion, which modifies the lru (https://github.com/mirage/ocaml-dns/pull/195/files#diff-d37c1f89ba842221e3ea46e11f61d936R194), (see https://pqwy.github.io/lru/doc/lru/Lru/F/module-type-S/index.html#val-promote))

cfcs:
Cool, that makes sense.
So this commit changes the tests slightly (`get` on line 77 potentially influenced by `get` on line 75 etc?), but that seems fine.
LGTM :)

PR #199, opened 2019-11-13
provide statistics in dns-server.mirage

hannesm:
at some point, the `counter_metrics` should be provided elsewhere (in metrics itself?), but for now this should be fine. //cc @cfcs

### DIFF
--- a/mirage/server/dns_server_mirage.ml
+++ b/mirage/server/dns_server_mirage.ml
@@ -11,30 +11,70 @@ module Make (P : Mirage_clock.PCLOCK) (M : Mirage_clock.MCLOCK) (TIME : Mirage_t
 
   module T = S.TCPV4
 
+  let create ~f =
+    let data : (string, int) Hashtbl.t = Hashtbl.create 7 in
+    (fun x ->
+       let key = f x in
+       let cur = match Hashtbl.find_opt data key with None -> 0 | Some x -> x in
+       Hashtbl.replace data key (succ cur)),
+    (fun () ->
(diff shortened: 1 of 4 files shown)


mirage/server/dns_server_mirage.ml
```diff
@@ -76,20 +117,22 @@ module Make (P : Mirage_clock.PCLOCK) (M : Mirage_clock.MCLOCK) (TIME : Mirage_t
...
+            | Some `Keep -> inc `Tcp_cache_add ; inc `Tcp_keep ; tcp_out := Dns.IM.add ip flow !tcp_out ; None
             | Some `Notify soa -> Some (`Notify soa)
             | Some `Signed_notify soa -> Some (`Signed_notify soa)
             | None -> None
           in
           maybe_update_state t >>= fun () ->
           maybe_notify recv_task t now elapsed n' >>= fun () ->
           (match answer with
-           | None -> Log.warn (fun m -> m "empty answer") ; Lwt.return_unit
+           | None -> Lwt.return_unit
```
cfcs: would it make sense to track number of empty answers too? If it made sense at some point as a debug message for tcp+udp, maybe a metric would be good to track down errors in the future?

hannesm: that is a good suggestion, although this is better done in the core dns library: here we have no information about the contents of a packet -- i.e. short packets (<12 byte), or answers (bit 17 = 1) would and should not provoke any replies -- and thus would be wrongly counted). I removed the log message since it was misleading.

PR #200, opened 2019-11-17
Dns_server.handle_buf: expose (optional) key name used to authenticate an update

hannesm:
Dns_mirage_server.primary: include above key name and source IP in on_update

this is mostly useful for a primary server with a git backend to create sensible commit messages

/cc @cfcs

### DIFF
--- a/mirage/server/dns_server_mirage.mli
+++ b/mirage/server/dns_server_mirage.mli
@@ -8,9 +8,12 @@ module Make (P : Mirage_clock.PCLOCK) (M : Mirage_clock.MCLOCK) (T : Mirage_time
                 Dns_server.Primary.s ->
                 (Dns_trie.t * ([ `raw ] Domain_name.t * Dns.Dnskey.t) list) option Lwt.t) ->
     ?timer:int -> ?port:int -> S.t -> Dns_server.Primary.s -> unit
-  (** [primary ~on_update ~timer ~port stack primary] starts a primary server on [port]
-     (default 53, both TCP and UDP) with the given [primary] configuration. [timer] is the
-     DNS notify timer in seconds, and defaults to 2 seconds. *)
+  (** [primary ~on_update ~timer ~port stack primary] starts a primary server on
+     [port] (default 53, both TCP and UDP) with the given [primary]
+     configuration. [timer] is the DNS notify timer in seconds, and defaults to
+     2 seconds. [on_update ~old key ip s] is a callback if the data served by
+     the primary server got updated by a potentially authenticated nsupdate
+     packet, the used [key] and source [ip] are passed to the callback. *)
 
   val secondary :
     ?on_update:(old:Dns_trie.t -> Dns_server.Secondary.s -> unit Lwt.t) ->
--- a/mirage/server/dns_server_mirage.mli
+++ b/mirage/server/dns_server_mirage.mli
@@ -2,7 +2,8 @@
 
 module Make (P : Mirage_clock.PCLOCK) (M : Mirage_clock.MCLOCK) (T : Mirage_time.S) (S : Mirage_stack.V4) : sig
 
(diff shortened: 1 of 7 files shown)


hannesm:
the third commit allows update packets with changes spanning over multiple zones. this is useful for several services, and it is unclear on the client side how to figure out where a zone would start / should be used.

mirage/server/dns_server_mirage.mli
```diff
@@ -2,14 +2,18 @@
 
 module Make (P : Mirage_clock.PCLOCK) (M : Mirage_clock.MCLOCK) (T : Mirage_time.S) (S : Mirage_stack.V4) : sig
 
-  val primary : ?on_update:(old:Dns_trie.t -> Dns_server.Primary.s -> unit Lwt.t) ->
+  val primary :
+    ?on_update:(old:Dns_trie.t -> [`raw] Domain_name.t option -> Ipaddr.V4.t -> Dns_server.Primary.s -> unit Lwt.t) ->
```
avsm: Just curious about the style here -- any reason not to label all the arguments so it's clear what (e.g.) the Ipaddr.V4.t represents?  I realise that conventionally we use labels to distinguish similar types for different uses, but in this case it's not clear what some of these represent without inspecting the code more closely.  However, this is a broader comment and not a blocker to this PR, since you are just extending the existing style by adding an argument.

mirage/server/dns_server_mirage.mli
```diff
@@ -2,14 +2,18 @@
...
                 (Dns_trie.t * ([ `raw ] Domain_name.t * Dns.Dnskey.t) list) option Lwt.t) ->
     ?timer:int -> ?port:int -> S.t -> Dns_server.Primary.s -> unit
-  (** [primary ~on_update ~timer ~port stack primary] starts a primary server on [port]
-     (default 53, both TCP and UDP) with the given [primary] configuration. [timer] is the
-     DNS notify timer in seconds, and defaults to 2 seconds. *)
+  (** [primary ~on_update ~timer ~port stack primary] starts a primary server on
+     [port] (default 53, both TCP and UDP) with the given [primary]
+     configuration. [timer] is the DNS notify timer in seconds, and defaults to
+     2 seconds. [on_update ~old key ip s] is a callback if the data served by
+     the primary server got updated by a potentially authenticated nsupdate
```
avsm:
```suggestion
     the primary server [s] got updated by a potentially authenticated nsupdate
```

server/dns_server.ml
```diff
@@ -671,32 +679,34 @@ let update_data trie zone (prereq, update) =
...
+        acc >>= fun (trie', zones) ->
+        match Dns_trie.lookup zone Soa trie, Dns_trie.lookup zone Soa trie' with
+        | Ok oldsoa, Ok soa when Soa.newer ~old:oldsoa soa ->
+          Ok (trie', (zone, soa) :: zones)
+        | _, Ok soa ->
+          let soa = { soa with Soa.serial = Int32.succ soa.Soa.serial } in
+          let trie'' = Dns_trie.insert zone Soa soa trie' in
+          Ok (trie'', (zone, soa) :: zones)
+        | Ok oldsoa, Error _ ->
+          (* zone removal!? *)
```
avsm: Is this worth a debug log here, as it seems an unusual case to encounter in the wild?

hannesm: depends on deployment etc. -- atm that's a bit untested (and not advertised) that you can dynamically add/remove zones. once that is tested (and used and deployed), I'm keen on getting rid of this comment, which imho serves its purpose for now.

mirage/server/dns_server_mirage.mli
hannesm: agreed -- I added labels in 169dee7

server/dns_server.ml
avsm: Sounds good

PR #204, opened 2020-01-08
Client API improvements

hannesm:
while the `` `Msg of string `` is sufficient for application usage, other components may want to react differently depending on the concrete error. This PR introduces a variant type for `getaddrinfo` for that (which we may want to extend for all possible errors, i.e. connection/truncation/...), if it turns out to be a good idea.

Additionally, this PR fixes an issue when a (valid, nodata) answer is processed by `follow_cname`, including a test -- instead of error'ing out, we handle CNAME replies with NoData (see https://tools.ietf.org/html/rfc2308 section 2.2) -- take a look at the corresponding test.

### DIFF
--- a/client/dns_client.ml
+++ b/client/dns_client.ml
@@ -185,8 +185,10 @@ struct
   (* result-bind-and-lift *)
   let (>>|=) a f = a >>| fun b -> Transport.lift (f b)
 
-  let getaddrinfo (type requested) t ?nameserver (query_type:requested Dns.Rr_map.key) name
-    : (requested, [> `Msg of string]) result Transport.io =
+  let get_resource_record (type requested) t ?nameserver (query_type:requested Dns.Rr_map.key) name
+    : (requested, [> `Msg of string
+                  | `No_data of [ `raw ] Domain_name.t * Dns.Soa.t
+                  | `No_domain of [ `raw ] Domain_name.t * Dns.Soa.t ]) result Transport.io =
     let domain_name = Domain_name.raw name in
     match Dns_cache.get t.cache (t.clock ()) domain_name query_type with
     | Ok `Entry (B (query_type', value)) ->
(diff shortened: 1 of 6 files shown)


client/dns_client.mli
```diff
@@ -80,7 +80,11 @@ sig
...
 
   val getaddrinfo : t -> ?nameserver:T.ns_addr -> 'response Dns.Rr_map.key ->
-    'a Domain_name.t -> ('response, [> `Msg of string ]) result T.io
+    'a Domain_name.t ->
+    ('response,
+     [> `Msg of string
+     | `No_data of [ `raw ] Domain_name.t * Dns.Soa.t
+     | `No_domain of [ `raw ] Domain_name.t * Dns.Soa.t ]) result T.io
   (** [getaddrinfo nameserver query_type name] is the [query_type]-dependent
       response from [nameserver] regarding [name], or an [Error _] message.
```
cfcs: mli-string needs to be updated to be kept in sync with return type.

client/dns_client.ml
```diff
@@ -229,27 +231,33 @@ struct
...
       in recv_loop Cstruct.empty) >>= fun r ->
       Transport.close socket >>= fun () ->
       Transport.lift r
 
+  let lift_cache_error m =
+    (match m with
+     | Ok a -> Ok a
+     | Error `Msg msg -> Error (`Msg msg)
+     | Error (#Dns_cache.entry as e) ->
+       Rresult.R.error_msgf "DNS cache error %a" Dns_cache.pp_entry e)
```
cfcs:
```suggestion
       Rresult.R.error_msgf "DNS cache error @[%a@]" Dns_cache.pp_entry e)
```

hannesm:
> for a lot of use-cases where the user (e.g. me) won't care

as I tried to elaborate in my initial comment: having a single `` `Msg of string `` is fine for some applications, but it turns out other use cases want to react on specific errors. I'm fine to develop this in a different function (though I don't like a `_expert` or `_complex` suffix).

cfcs:
@hannesm Yes, I agree with you that typed errors can be useful, and there are use-cases for both typed + stringly typed errors, the latter mostly being convenient for combining with other `result` things. I don't think there's anything wrong with the code in this PR, I just don't feel so good about cluttering the interface for people who will not care about typed errors, so I would prefer a separate function (or moving the current to a `_easy` function or something like that).

I also believe that if we return variant errors, we should provided a `error_msg_of_variant` function to convert to the stringly typed error, so people who only cares about one or two specific error codes can handle those, then resume to the "easy" mode.

hannesm:
I introduced in 55528e589d9f0dd6f553ea4ecfbc7a4283a4c5fe a `get_resource_record` function. hope this addresses your concerns. I am not yet providing a `pp_error`, since I first would like to figure out more detailed what should be in the variant (and whether a `` `Msg of string `` is acceptable). Once I figured this out, I'd be happy to push to this PR and merge eventually.

PR #207, opened 2020-01-12
server bugfix (in update); metrics

hannesm:
this adds more tests, esp. for the case a server has multiple zones and only one is updated. in 4.3.0 this lead to SOA serial bumps of all zones the primary is serving.

another test is a single nsupdate which updates multiple zones at once.

in dns server, also expose some counter metrics about received and transmitted dns packets

### DIFF
--- a/server/dns_server.ml
+++ b/server/dns_server.ml
@@ -718,24 +718,39 @@ let update_data trie zone (prereq, update) =
     trie, []
   else
     let zones =
-      (* accumulate old and new zones *)
-      let one t s =
-        Dns_trie.fold Rr_map.Soa t (fun z _ s -> Domain_name.Set.add z s) s
-      in
-      one trie' (one trie Domain_name.Set.empty)
+      (* figure out the zones where changes happened *)
+      (* for each element in the map of updates (domain_name -> update list),
+         figure out the zone of the domain_name. Since zone addition and
+         removal is supported, this name may be present in both the old trie and
+         the new trie' (update), only in the old trie (delete), only in the new
+         trie (add). *)
+      Domain_name.Map.fold (fun name _ acc ->
+          match Dns_trie.zone name trie, Dns_trie.zone name trie' with
+          | Ok (z, _), _ | _, Ok (z, _) -> Domain_name.Set.add z acc
+          | Error e, Error _ ->
+            Log.err (fun m -> m "couldn't find zone for %a in either trie: %a"
+                        Domain_name.pp name Dns_trie.pp_e e);
+            acc) update Domain_name.Set.empty
     in
+    (* now, for each modified zone, ensure the serial in the SOA increased, and
+       output the zone name and its zone. *)
     Domain_name.Set.fold (fun zone (trie', zones) ->
         match Dns_trie.lookup zone Soa trie, Dns_trie.lookup zone Soa trie' with
         | Ok oldsoa, Ok soa when Soa.newer ~old:oldsoa soa ->
+          (* serial is already increased in trie', nothing to do *)
           trie', (zone, soa) :: zones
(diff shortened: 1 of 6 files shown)


server/dns_server.ml
```diff
@@ -686,10 +719,14 @@ let update_data trie zone (prereq, update) =
...
     let zones =
       (* accumulate old and new zones *)
-      let one t s =
-        Dns_trie.fold Rr_map.Soa t (fun z _ s -> Domain_name.Set.add z s) s
-      in
-      one trie' (one trie Domain_name.Set.empty)
+      Domain_name.Map.fold (fun name _ acc ->
+          match Dns_trie.zone name trie', Dns_trie.zone name trie with
+          | Ok (z, _), _ -> Domain_name.Set.add z acc
+          | _, Ok (z, _) -> Domain_name.Set.add z acc
```
cfcs: 1) I don't really understand what's going on here, but in line 733 there's:
    ```ocaml
          | Ok oldsoa, Ok soa when Soa.newer ~old:oldsoa soa ->
  The `z`from `trie` doesn't get added when `Dns_trie.zone name trie'` works, tough, so the `oldsoa` will presumably come from `trie'` - will that be a problem, or are they guaranteed to be identical or something like that in the `Ok oldsoa, Ok soa` case?

2) It would be nice if there was consistency in the matches so `trie'` is always before/after `trie`, instead of random ordering. It's hard enough to spot the `'` in the first place. :)

server/dns_server.ml
```diff
@@ -686,10 +719,14 @@ let update_data trie zone (prereq, update) =
...
+          | Error e, _ ->
```
cfcs:
```suggestion
          | Error e, Error _ ->
```
To make it clear that this case only deals with two errors.

test/server.ml
```diff
@@ -1084,6 +1084,103 @@ module A = struct
...
+    let key = n_of_s "foo._update.one.com" in
+    let soa' = { soa with serial = Int32.succ soa.serial } in
+    let trie' =
+      Dns_trie.insert (n_of_s "one.com") Rr_map.Soa soa'
+        (Dns_trie.insert foo entry_key entry_val server'.data)
+    in
+    Alcotest.(check (result (pair trie_test zone_test) rcode_test)
+                __LOC__ (Ok (trie', [ n_of_s "one.com", soa' ]))
+                (handle_update server' `Udp (Some key) q up));
+    let server'' = Dns_server.with_data server' trie' in
```
cfcs: maybe move `server''` bind down below `old_trie` to make it clearer that `old_trie` is not using `server'` by mistake?

test/server.ml
```diff
@@ -1084,6 +1084,103 @@ module A = struct
...
+      Domain_name.Map.empty,
+      Domain_name.Map.singleton foo [ Packet.Update.Remove (Rr_map.K entry_key) ]
+    in
+    Alcotest.(check (result (pair trie_test zone_test) rcode_test)
+                __LOC__ (Ok (old_trie, [ n_of_s "one.com", soa'' ]))
+                (handle_update server'' `Udp (Some key) q up'));
+    let foo_two = n_of_s "foo.two.com" in
+    let key' = n_of_s "foo._update.two.com" in
+    let q' = Packet.Question.create (n_of_s "two.com") Soa in
+    let up =
```
cfcs: This is not allowed, you cannot use both `up` and `up'` and then revert to shadowing `up` ;-)
(maybe `up_foo_two` would be a better name?)

- same below where `q'` and `up` are shadowed again, just above the definition of `server''''`.

hannesm:
thanks for the review, I think I addressed all your comments adequately.

PR #209, opened 2020-01-28
Stub resolver

hannesm:
this is a series of commits, mainly implementing a stub resolver using in front a primary Dns_server (local overrides), another server for the reserved entries (example.com, various private IP ranges), and a forwarding caching resolver using the Dns_client API via TCP.

Other noteworthy improvements:
- Dns now provides a Metrics source generator (counter_metrics), used by server, cache, stub-resolver
- Dns_tsig.encode_and_sign has an optional `?mac:Cstruct.t` argument (to authenticate replies to authenticated requests)
- Dns_server.create is now public API

//cc @cfcs


mirage/stub/dns_stub_mirage.ml
```diff
@@ -0,0 +1,401 @@
...
+          Log.info (fun m -> m "end of file reading from resolver");
+          cancel_all t;
+          Lwt.return_unit
+        | Ok (`Data cs) ->
+          let rec handle_data data =
+            let cs_len = Cstruct.len data in
+            if cs_len > 2 then
+              let len = Cstruct.BE.get_uint16 data 0 in
+              if cs_len - 2 >= len then
+                let packet, rest = Cstruct.split data (len + 2) in
```
cfcs: this will always allocate two new buffers, would it make sense to have something like
```ocaml
let packet, rest =
  if cs_len - 2 = len then packet, Cstruct.empty
  else Cstruct.split data (len + 2)
in
```
to prevent the extra allocations in the case where there's no trailing data? That does mean that we're passing on a mutable reference to `packet`, but as far as I can tell we don't use `data` again in that case.

mirage/stub/dns_stub_mirage.ml
```diff
@@ -0,0 +1,401 @@
...
+                        S.TCPV4.pp_error e) ;
+            Error (`Msg "connect failure")
+          | Ok flow ->
+            inc `Recursive_connections;
+            Lwt.async (fun () -> read_loop t flow);
+            t.flow <- Some flow;
+            Ok ({ t ; id = 0 })
+
+      let close _f =
+        (* ignoring this here *)
```
cfcs: shouldn't this issue `cancel_all` and close the flow?

mirage/stub/dns_stub_mirage.ml
```diff
@@ -0,0 +1,401 @@
...
+        t.requests <- IM.remove id t.requests;
+        match data with
+        | Ok cs -> Ok cs
+        | Error `Msg m -> Error (`Msg m)
+
+      let send f s =
+        match f.t.flow with
+        | None -> Lwt.return (Error (`Msg "no connection to resolver"))
+        | Some flow ->
+          let id = Cstruct.BE.get_uint16 s 2 in
```
cfcs: Is this always safe, ie do we need a guard on the length?

mirage/stub/dns_stub_mirage.ml
```diff
@@ -0,0 +1,401 @@
...
+      inc `Authoritative_answers;
+      Some (build_reply ?additional (`Answer answer))
+    | Error (Rcode.NotAuth, _) -> None
+    | Error (rcode, answer) ->
+      inc `Authoritative_errors;
+      let data = `Rcode_error (rcode, Packet.opcode_data data, answer) in
+      Some (build_reply ?additional:None data)
+
+  let tsig_decode_sign server proto packet buf build_reply =
+    match Dns_server.handle_tsig server (Ptime.v (P.now_d_ps ())) packet buf with
```
cfcs: Two points:
- seems to me that `now` could be reused since there are no Lwt yield points here
- should we allow overriding this time with a `?time=(Ptime.v (P.now_d_ps ()))` parameter for the purpose of tests?

mirage/stub/dns_stub_mirage.ml
```diff
@@ -0,0 +1,401 @@
...
+      server t proto ip packet data build_reply >>= function
+      | Some data -> Lwt.return (Some data)
+      | None ->
+        (* next look in reserved trie! *)
+        let question, data = packet.Packet.question, packet.Packet.data in
+        match query_server t.reserved question data build_reply with
+        | Some data -> inc `Reserved_answers ; Lwt.return (Some data)
+        | None -> resolve t packet.Packet.question packet.Packet.data build_reply
+
+  let create ?(nameserver = Ipaddr.V4.of_string_exn "141.1.1.1") ?(size = 10000) ?(on_update = fun ~old:_ ?authenticated_key:_ ~update_source:_ _trie -> Lwt.return_unit) primary stack =
```
cfcs: what is this nameserver, could we document that in a comment and/or mli eventually?

cfcs: what motivates the `?size=10000` default, why not e.g. 65000 ?

src/dns.mli
```diff
@@ -1126,3 +1126,7 @@ module Tsig_op : sig
   val no_sign : sign
   (** [no_sign] always returns [None]. *)
 end
+
+(**/**)
```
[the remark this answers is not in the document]
hannesm: that everything below won't end up in the generated API documentation. I do think there is a better place for `counter_metrics` in the ocaml ecosystem, so it should not be used by other libraries...

client/dns_client.ml
```diff
@@ -26,6 +26,14 @@ module Pure = struct
...
 
   (* name: the originally requested domain name. *)
+  (* NOTE that this function compresses answers:
+     foo.example CNAME 500 bar.example
+     bar.example A 300 1.2.3.4
+     is compressed to:
+     foo.example A 300 1.2.3.4
+     -> which is fine for applications (i think so)
+     -> which is struggling for the cache (not entirely sure about this tbh)
+     -> it is not clear whether it meets the DNS specifications nicely *)
```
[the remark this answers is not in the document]
hannesm: after re-thinking, I guess for "good answers" nobody cares. but for "no such domain" / "no data" answers (former is "this domain name does not exist", latter "this resource record type does not exist for this domain name -- see https://tools.ietf.org/html/rfc2308 for error handling), this may make a difference since the client gets "the wrong SOA" -- i.e. an unexpected authority (where the requested domain name does not match) - which strict clients may want to drop on the floor instead of accepting it.. I have no concrete example we can test it on, but I'm sure we can find some on the internet. :)

tsig/dns_tsig.mli
```diff
@@ -21,9 +21,9 @@ type s = [ `Key_algorithm of Dnskey.t | `Tsig_creation | `Sign ]
 val pp_s : s Fmt.t
 (** [pp_s ppf s] pretty-prints [s] on [ppf]. *)
 
-val encode_and_sign : ?proto:proto -> Packet.t -> Ptime.t -> Dns.Dnskey.t ->
-  'a Domain_name.t -> (Cstruct.t * Cstruct.t, s) result
-(** [encode_and_sign ~proto t now dnskey name] signs and encodes the DNS
+val encode_and_sign : ?proto:proto -> ?mac:Cstruct.t -> Packet.t -> Ptime.t ->
+  Dns.Dnskey.t -> 'a Domain_name.t -> (Cstruct.t * Cstruct.t, s) result
+(** [encode_and_sign ~proto ~mac t now dnskey name] signs and encodes the DNS
     packet. *)
```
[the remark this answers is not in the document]
hannesm: the `~mac` - as mentioned in https://tools.ietf.org/html/rfc2845 Section 4 - is the message authentication code of the request, which is part of the things to be signed in the response.

return value of this function is in the left case a pair of "encoded packet" and "resulting mac". the former to be send out, the latter to be preserved (for verifying a reply if this was a request).

mirage/stub/dns_stub_mirage.ml
```diff
@@ -0,0 +1,401 @@
...
+        nameserver : ns_addr ;
+        stack : stack ;
+        mutable flow : S.TCPV4.flow option ;
+        mutable requests : (Cstruct.t, [ `Msg of string ]) result Lwt_condition.t IM.t ;
+      }
+      type flow = { t : t ; mutable id : int }
+
+      let create
+          ?rng
+          ?(nameserver = `TCP, (Ipaddr.V4.of_string_exn "91.239.100.100", 53))
```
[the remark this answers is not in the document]
hannesm: that sounds like an excellent idea we should investigate further to avoid the duplications (maybe open a separate issue, since this is not in scope for this PR)?

mirage/stub/dns_stub_mirage.ml
hannesm: close is called by the dns_client logic, no? and there it assumes that each connection is used by a single request only -- while here we multiplex across multiple requests / responses over a single connection, and better keep that connection alive instead of closing it upon a dns client logic "this request was finished"...

mirage/stub/dns_stub_mirage.ml
hannesm: 141.1.1.1 is an IPv4 address I remember -- an anycast recursive resolver run by cable&wireless. it should not be here (but iirc i encountered some issues connecting to censurfridns from marrakesh, so I put another nameserver in here) -- I agree with your above comment that we should have only a single IP hardcoded once :)

size was chosen arbitrarily... this is the number of resource records in the cache (while for use cases, I'd be more interested in being able to specify "number of bytes used by the cache").

client/dns_client.ml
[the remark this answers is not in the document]
cfcs: I also wonder how this plays in with DNSSEC - if I request `www.foo.example` I would intuitively expect there to be a signed path leading to the answer of `www.bar.example has A record 1.2.3.4` (signed by bar.example?), in that case including the CNAME might make sense?

tsig/dns_tsig.mli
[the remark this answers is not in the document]
cfcs: I think both the RFC reference and this explanation of the returned tuple would make sense to include in the mli :)

mirage/stub/dns_stub_mirage.ml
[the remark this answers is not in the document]
cfcs: Opened #210

mirage/stub/dns_stub_mirage.ml
cfcs: Ah, I see. So instead here we could implement logic to reset the timeout for the shared connection?

hannesm: yes, we should discuss connections (setup & teardown & timeouts) soon (maybe in our tomorrow before noon session).

mirage/stub/dns_stub_mirage.ml
hannesm: fixed the first point; the second I'm not sure about -- developing a testsuite should lead to a refactoring anyways and hopefully establishing a (lwt-free) dns-stub package (atm it's all in dns-stub.mirage).

mirage/stub/dns_stub_mirage.ml
hannesm: yes, everything ocaml-dns creates is at least 12 bytes long

mirage/stub/dns_stub_mirage.ml
cfcs: fine with me :)

cache/dns_cache.ml
```diff
@@ -177,21 +177,21 @@ let update_ttl entry ~created ~now =
...
+    | Error e -> metrics `Drop; Error e
 
 (* XXX: we may want to define a minimum as well (5 minutes? 30 minutes?
    use SOA expiry?) MS used to use 24 hours in internet explorer
 
 from RFC1034 on this topic:
 The idea is that if cached data is known to come from a particular zone,
 and if an authoritative copy of the zone's SOA is obtained, and if the
-zone's SERIAL has not changed since the data was cached, then the TTL of
+zone's SERIAL has not changed smetricse the data was cached, then the TTL of
```
cfcs:
```suggestion
zone's SERIAL has not changed since the data was cached, then the TTL of
```

PR #212, opened 2020-01-31
cache: clip ttl to hour..ttl..week

hannesm:
this is likely controversial -- reasoning behind this is https://00f.net/2019/11/03/stop-using-low-dns-ttls/ and the observation that e.g. github.com has ridiciously low ttls for no good reason... this PR is mostly a reminder that the current behaviour is not optimal for most setups, and should be refined. an option could be to provide configuration arguments (though dns cache doesn't have configuration atm); or we decide to merge and use the increased lower ttl. i tried this at the reproducible builds summit in december 2019 in marrakesh and didn't hear any complaints -- i suspect nobody noticed...

### DIFF
--- a/cache/dns_cache.ml
+++ b/cache/dns_cache.ml
@@ -211,15 +211,21 @@ and 1035 6.2:
    maximum value! *)
 let week = Int32.of_int Duration.(to_sec (of_day 7))
 
-let clip_ttl_to_week entry =
+let min_ttl = Int32.of_int (Duration.(to_sec (of_hour 1)))
+
+let clip_ttl_to_min_hour_max_week entry =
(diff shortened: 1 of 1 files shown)


cfcs:
I think it's controversial, it will make all users of the cache (e.g. the client?) useless for a lot of use-cases, for instance when taking part in a Consul network (where low TTL is desired) etc.

Seems to me that this should be something configurable in a resolver (and maybe also client?) rather than hardcoded.

hannesm:
I agree this should be somehow configurable, and in some scenarios it is a good idea, while it is not such a good idea in other scenarios. let's keep this PR open for now (it is as well not urgent to get the clip in - but otherwise we'll forget about it).

Issue #213, opened 2020-02-03
Glue record for primary DNS server


dinosaure:
Currently, I used `ocaml-dns` as my primary DNS server (`primary-git`) and use `ns6.gandi.net` as my secondary DNS server. However. if I do nothing (eg. just fill my zone file and launch the DNS server), I got an error:

```
[dns_server] lookup for A ns6.gandi.net returned nothing as well
```

@hannesm gaves me a solution where:
1) put a new entry in my zone file like `ns6.gandi.net. A X.X.X.X`
2) make a new `gandi.net` zone file with `ns6.gandi.net. A X.X.X.X`

does not work:
1) the server told me: `error during loading git an entry of x25519.net is not in its zone, won't handle this`
2) the server told me: `error during loading git text: couldn't find zone gandi.net: not authoritative`

 Currently the solution is to make a new `gandi.net` zone file, write an `SOA` and fill it with `ns6.gandi.net`. It's work but may be it's not the proper way to do that.

hannesm:
thanks for your report.

I implemented a solution for this (putting the ns6.gandi.net address record in your x25519.net zone). the downside is that now the zone file may include non-authoritative data, which needs to be updated (whenever ns6.gandi.net changes their IP address - shouldn't happen too frequently ;)). (see the linked PRs to this repository and roburio/unikernels). please let me know if this works fine for you.

an alternative implementation could use a dns-client to resolve ns6.gandi.net, but I dislike the dependency of the primary authoritative nameserver to a dns-client/resolver.

PR #217, opened 2020-02-05
Find secondaries

hannesm:
this strengthens the `lookup` function (to return `` `NotAuthoritative`` more often), and relaxes the `Notification.secondaries` to also find glue records in the trie. This addresses (with some further changes to the primary-git unikernel #213)

also add tests for lookup in a trie where the zone does not contain a SOA record (glue-only), `lookup_glue` returns the resource record set

//cc @cfcs @dinosaure

### DIFF
--- a/server/dns_server.ml
+++ b/server/dns_server.ml
@@ -466,10 +466,9 @@ module Notification = struct
         | Error _ -> ns
         | Ok prim -> Domain_name.Host_set.remove prim ns
       in
-      (* TODO AAAA records / use lookup_glue? *)
       Domain_name.Host_set.fold (fun ns acc ->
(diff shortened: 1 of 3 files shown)


server/dns_server.ml
```diff
@@ -466,13 +466,12 @@ module Notification = struct
         | Error _ -> ns
         | Ok prim -> Domain_name.Host_set.remove prim ns
       in
-      (* TODO AAAA records / use lookup_glue? *)
```
cfcs: we still don't handle `AAAA` here, correct?

hannesm: true, i pushed another commit to retain the TODO (but remove the `use lookup_glue`)

PR #219, opened 2020-02-19
revise dns_certify with new ACME / let's encrypt in mind

hannesm:
the main change is that let's encrypt now very conveniently provides the chain of trust as well as the end-entity certificate. this means that we no longer need to carry the hard-coded certificate around. but this implies that we need to store the CA chain somewhere (again, as TLSA records :)

I'm still not entirely happy with the dns_certify API, but I think this is moving into the right direction.

### DIFF
--- a/certify/dns_certify.ml
+++ b/certify/dns_certify.ml
@@ -1,5 +1,44 @@
 open Dns
 
+let tlsa_is usage sel typ t =
+  t.Tlsa.cert_usage = usage &&
+  t.Tlsa.selector = sel &&
+  t.Tlsa.matching_type = typ
+
+let is_csr t =
+  tlsa_is Tlsa.Domain_issued_certificate Tlsa.Private Tlsa.No_hash t
+
+let csr req =
+  let data = X509.Signing_request.encode_der req in
+  {
+    Tlsa.matching_type = Tlsa.No_hash ;
+    cert_usage = Tlsa.Domain_issued_certificate ;
+    selector = Tlsa.Private ;
+    data
+  }
+
+let is_certificate t =
+  tlsa_is Tlsa.Domain_issued_certificate Tlsa.Full_certificate Tlsa.No_hash t
+
+let certificate cert =
+  let data = X509.Certificate.encode_der cert in
+  {
+    Tlsa.matching_type = Tlsa.No_hash ;
+    cert_usage = Tlsa.Domain_issued_certificate ;
+    selector = Tlsa.Full_certificate ;
+    data
+  }
+
+let is_ca_certificate t =
+  tlsa_is Tlsa.CA_constraint Tlsa.Full_certificate Tlsa.No_hash t
+
+let ca_certificate data = {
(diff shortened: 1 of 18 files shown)


hannesm:
once this is in and tested successfully, I'd like to go ahead and draft a release unless there are objections

certify/dns_certify.ml
```diff
@@ -83,26 +126,35 @@ let pp_q_err ppf = function
...
 
+let tlsas_to_certchain public_key tlsas =
+  let certificates, ca_certificates =
+    Rr_map.Tlsa_set.fold (fun tlsa (certs, cacerts as acc) ->
+        if is_certificate tlsa || is_ca_certificate tlsa then
+          match X509.Certificate.decode_der tlsa.Tlsa.data with
+          | Ok cert ->
+            if is_certificate tlsa then
+              (cert :: certs, cacerts)
+            else (* must be is_ca_certificate *)
```
cfcs: does this comment suggest we should check for `is_ca_certificate` ?

certify/dns_certify.ml
```diff
@@ -83,26 +126,35 @@ let pp_q_err ppf = function
...
+              (certs, cert :: cacerts)
+          | _ -> acc
```
cfcs: here we ignore the parsing `Error _` ?

certify/dns_certify.ml
hannesm: that's checked some lines above.. this code will only ever be executed if either `is_certificate` or `is_ca_certificate` is true.. checking it here again means I'd need to handle the impossible case of neither being true, which I'd like to avoid (but the code may be refactored for good to avoid these cases)

certify/dns_certify.ml
hannesm: yes, this is a shortcoming of the code (but the same behaviour as previously). maybe best to emit a log message..

certify/dns_certify.ml
cfcs: I think it'd be good if the comment could explain that (I was a bit unsure if `must be` meant "we can logically derive that it is because we just checked" or "it really really should be, but we don't know if it is")

certify/dns_certify.mli
```diff
@@ -13,6 +13,30 @@ val letsencrypt_name : 'a Domain_name.t ->
 (** [letsencrypt_name host] is the service name at which we store let's encrypt
     certificates for the [host]. *)
 
+val is_csr : Dns.Tlsa.t -> bool
+(** [is_csr tlsa] is true if [tlsa] is a certificate signing request of
+    interest for this library. *)
```
cfcs: What does "of interest for this library" mean here?

hannesm: "this library" is dns-certify, which uses TLSA records for certificates.

cfcs: Right, but what makes a certificate signing request interesting/not interesting to the library?

PR #221, opened 2020-03-28
client: stick to RFC 6761 (special purpose domain names)

### LINKED ISSUE #220: Client does not special-case `localhost`, `ip6-localhost`, `ip6-loopback`
On Debian the default `/etc/hosts` file ships with:
```
127.0.0.1	localhost
::1		localhost ip6-localhost ip6-loopback
ff02::1		ip6-allnodes
ff02::2		ip6-allrouters
```

It appears users are confused by `dns-client` trying to resolve at least `localhost` upstream. We should consider if we want to support at least `localhost`, and perhaps the others. Which then begs the question of whether we should have a `/etc/hosts`-equivalent in the library that users are able to override, e.g. this could be part of the [`S.create`](https://github.com/mirage/ocaml-dns/blob/master/client/dns_client.mli#L47) API.

Where to put it: Intuitively I'd say the cache would seem like a good place?

For anecdotal real-world evidence of this being confusing to users: https://github.com/roburio/openvpn/issues/28

hannesm:
handle localhost and invalid appropriately as 'Name Resolution APIs and Libraries'

more specifically, `*.localhost.` resolves to A 127.0.0.1 AAAA ::1, all other query types NXDomain
`*.invalid` resolves to a negative reply (NXDomain).

fixes #220 //cc @cfcs

### DIFF
--- a/client/dns_client.ml
+++ b/client/dns_client.ml
@@ -167,6 +167,24 @@ module type S = sig
   val lift : 'a -> 'a io
 end
 
+let localhost = Domain_name.of_string_exn "localhost"
+let localsoa = Soa.create (Domain_name.prepend_label_exn localhost "ns")
+let invalid = Domain_name.of_string_exn "invalid"
+let invalidsoa = Soa.create (Domain_name.prepend_label_exn invalid "ns")
+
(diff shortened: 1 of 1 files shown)


hannesm:
better viewed without disturbing whitespace changes by [following this link](https://github.com/mirage/ocaml-dns/pull/221/files?w=1)

PR #235, opened 2020-07-28
dns: fix #234: edns flags are 16 bit (not 32), dnssec_ok is the topmost

### LINKED ISSUE #234: Error on 32bit architecture
I'm not really aware about DNS but the CI got an error about an integer which is larger than what is authorized with 31 bits here: https://github.com/mirage/ocaml-dns/blob/50fe8eef13dd880fb98e84a028440f09bc488827/src/dns.ml#L1545

However, I'm not sure about this line where `flags` seems to be a 16 bits integer, so `dnssec_ok` will be always equal to `false`.

hannesm:


dinosaure:
The error still exist here:

https://github.com/mirage/ocaml-dns/blob/50fe8eef13dd880fb98e84a028440f09bc488827/src/dns.ml#L1573

I hope that it's the last one.

hannesm:
this is now ready for review :) 32bit CI no longer fails

avsm:
Glad to see that CI mode is catching a few things :-)

PR #240, opened 2020-10-14
dns-client.resolvconf: a (ocamllex + ocamlyacc) parser for resolv.conf

hannesm:
On Unix systems, it is good to just re-use whichever nameservers are configured in resolv.conf. There's no tight specification of resolv.conf, but here is a parser influenced by [rusts resolv-conf](https://github.com/tailhook/resolv-conf).

The error reporting is not optimal, and most fields are ignored (only `nameserver` entries are taken into consideration). For basic / common usage this is sufficient.


dinosaure (approved):
Completely agree with this PR!

PR #245, opened 2020-11-05
DNS: revise TXT resource record encoding and storage

### LINKED ISSUE #244: zone parser and TXT records
individual TXT entries are limited to 255 characters (due to a single byte defining their length). this is enforced by the zone parser.

but, e.g. dkim keys are longer, and it would be very convenient to put them into a single TXT record --> ocaml-dns should split them internally into multiple TXT records.

hannesm:
Previously, each TXT record hold a single string up to 255 characters in length.
When the length was exceeded, a new resource record was used. This is wasteful
(name, type, class, len needs to be repeated) and unexpected in some cases
(the requesting party gets multiple resource records which content needs to be
concatenated). The Txt.encode and Txt.decode did not behave in the same way
(decode split individual character strings, encode encoded only a single one).

When I tried to introduce some DKIM keys into my domains, I hit this issue that
DKIM check tools couldn't handle the responses from my servers. With this change
(and regression test):
(a) the Txt record can hold values that exceed 255 bytes
(b) in Txt.decode, a single resource record is a single entry in the map (the
    charstrings are concatenated)
(c) Txt.encode deals with values that exceed 255 bytes (by chunking them into
    255 byte long pieces, followed by a single byte length field, ...
(d) the zone parser is adapted to accept charstrings of length up to 65279
    bytes in a TXT entry (the rdlen is 16 bit (= 2 ^ 16), every 255 bytes we
    have to use 256)

This fixes #244.

### DIFF
--- a/src/dns.ml
+++ b/src/dns.ml
@@ -1084,13 +1084,26 @@ module Txt = struct
         more (d::acc) off
     in
     let txts = more [] 0 in
-    Ok (txts, names, off + len)
+    Ok (String.concat "" txts, names, off + len)
 
   let encode txt names buf off =
-    let len = String.length txt in
-    Cstruct.set_uint8 buf off len ;
-    Cstruct.blit_from_string txt 0 buf (succ off) len ;
-    names, off + len + 1
+    let rec more off txt =
+      if txt = "" then
+        off
+      else
+        let len = String.length txt in
+        let len, rest =
+          if len > 255 then
+            255, String.(sub txt 255 (len - 255))
+          else
+            len, ""
+        in
+        Cstruct.set_uint8 buf off len ;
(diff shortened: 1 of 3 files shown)


hannesm:
This is live on ns0.robur.coop and ns1.robur.coop (with RR `TXT mail._domainkey.robur.coop`), and ready for review. One question I have is whether anyone knows of applications that require specific (user-controlled) splitting of TXT records, i.e. is the current behaviour fine to concatenate a (zone file entry)  `yyy TXT "abc" "def"` to `yyy TXT "abcdef"`?

hannesm:
So the above question is whether whatever a user puts in a zone file in respect to split of charstrings in a TXT record should be retained, or whether the `String.concat ""` is fine (and on the binary encoding site always chunk up 255 bytes until the last one).

PR #247, opened 2020-12-11
dns-certify-mirage: only require dnskey to be valid for hostname

hannesm:
previously, all additional hostnames needed to be present in there as well.
this is not needed: the CSR to be uploaded is only for the zone of hostname
(the let's encrypt thingy needs to ensure that it can access all zones in the
CSR)


hannesm:
this is aligned with the ocertify utility.

PR #249, opened 2020-12-21
dns-certify improvements and dual stack support

hannesm:
this includes improvements for dns-certify:
- allow ``[`raw] Domain_name.t`` for additional hostnames
- avoid dependency on tls, return `X509.Certificate.t list * Mirage_crypto_pk.RSA.priv` instead of `Tls.Config.own_cert` (makes this more composable as well)
- rebased dual stack patches from #239

### DIFF
--- a/app/ocertify.ml
+++ b/app/ocertify.ml
@@ -151,7 +151,7 @@ let hostname =
 
 let more_hostnames =
   let doc = "Additional hostnames to be included in the certificate as SubjectAlternativeName extension" in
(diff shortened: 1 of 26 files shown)


hannesm:
The path forward is to merge this (it was reviewed as #239 earlier), and continue with fixes and patches. Since this breaks the (mirage) API, the next release will be a major version.

hannesm:
since the previous main branch already contained various bugfixes, and the first commit in here is a bugfix as well, I prepared a `4` branch and cut a release with these (4.6.3) -- the next release will be 5 which features dual stack support and is API breaking.

Issue #215, opened 2020-02-04
DNS client "compresses" answers with CNAMEs


hannesm:
from https://github.com/mirage/ocaml-dns/pull/209#discussion_r372138890 

NOTE that this function compresses answers:
```
foo.example CNAME 500 bar.example
bar.example A 300 1.2.3.4
```
is compressed to:
```
foo.example A 300 1.2.3.4
```

- this is likely an issue once we support DNSSec
- this may be an issue for clients handling NoData / NXDomain (https://tools.ietf.org/html/rfc2308) and do in-bailiwick (see https://krva.blogspot.com/2008/10/bailiwick-in-dns-requests.html) scrubbing
- we should test and evaluate with clients

cfcs:
[uBlock requires](https://github.com/gorhill/uBlock/releases/tag/1.25.0) (original [issue here](https://github.com/uBlockOrigin/uBlock-issues/issues/780)) knowing the full list of CNAMEs (or at least the first/last). If someone wanted to write a similar application in OCaml, they would likely need this information to be available.

hannesm:
since #236 introduced `get_raw_reply` for dns client, there's no need to revise this design decision. imho for a dns client it makes perfect sense that cnames are followed.

PR #251, opened 2021-01-17
DNSSec (RFC 4034 et al) support

hannesm:
This is work in progress, some bits and pieces are still missing (such as NSEC RR, canonicalization of Domain_name.t and RRSet). Feedback welcome.

### DIFF
--- a/src/dns.mli
+++ b/src/dns.mli
@@ -836,7 +836,7 @@ module Rr_map : sig
   val with_ttl : 'a key -> 'a -> int32 -> 'a
   (** [with_ttl k v ttl] updates [ttl] in [v]. *)
 
-  val prep_for_sig : [`raw] Domain_name.t -> Rrsig.t -> t -> (Cstruct.t, [ `Msg of string ]) result
+  val prep_for_sig : [`raw] Domain_name.t -> Rrsig.t -> 'a key -> 'a -> (Cstruct.t, [ `Msg of string ]) result
 
 end
 
--- a/src/dns.mli
+++ b/src/dns.mli
@@ -444,6 +444,9 @@ end
 module Nsec3 : sig
   type f = [ `Opt_out ]
 
+  val flags_of_int : int -> f option
+  (** [flags_of_int v] is [f]. *)
+
   type t = {
     flags : f option ;
     iterations : int ;
@@ -838,7 +841,13 @@ module Rr_map : sig
   (** [ppk ppf k] pretty-prints [k]. *)
 
   val of_int : ?off:int -> int -> (k, [> `Malformed of int * string ]) result
-  (** [of_int ~off i] constructs a [k] of the  provided integer. *)
+  (** [of_int ~off i] constructs a [k] of the provided integer. *)
+
+  val to_int : 'a key -> int
+  (** [to_int k] is the integer representing the key [k]. *)
+
+  val of_string : string -> (k, [> `Msg of  string ]) result
+  (** [of_string i] constructs a [k] of the provided string. *)
 
   val names : 'a key -> 'a -> Domain_name.Host_set.t
   (** [names k v] are the referenced domain names in the given binding. *)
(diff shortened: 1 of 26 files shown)


hannesm:
- key tag computation missing
- CI shows failures on 32bit systems that should be resolved

src/dns.mli
```diff
@@ -341,19 +350,72 @@ module Dnskey : sig
...
+  val compare : t -> t -> int
+  (** [compare a b] compares the RRSIG [a] with [b]. *)
+end
+
+(** DS *)
+module Ds : sig
+  type digest_type =
+    | SHA1
+    | SHA256
+    | SHA384
```
avsm, line 398: is this missing GOST from RFC5933? I think that'll cover all the assigned ones

avsm:
This is a surprisingly small PR for such a big feature! Nice work.

src/dns.mli
hannesm: indeed (https://www.iana.org/assignments/ds-rr-types/ds-rr-types.xhtml) -- but my approach is to name only these we actually have support for (and all others for future compatibility fall into the `Unknown`). since I don't see a GOST implementation (neither do I know anyone who asked for it), I'll skip that for now.

avsm: Sounds good. My understanding is that GOST is on the way out anyway.

hannesm:
now rebased onto the main branch

hannesm:
(likely incomplete) leftovers:
- cname proofs (needs dns-client compress changes #215)
- nsec3 support: base32, hashing, proofs of non-existance, opt-in
- dealing with superfluous (and badly signed) nsec/nsec3 (embedded by some person in the middle)
- "aggressive caching" (resolver)
- warn on SHA1 usage
- server side (+support in zone reader)

new RFCs supported (more?)
- RFC 7129 (denial of existence)
- RFC 4034 (dnssec RR)
- RFC 4035 (dnssec protocol)
- RFC 4033 (dnssec base)
- RFC 5155 (nsec3) TODO (RR is there), nsec3param
- RFC 4648 (base32) TODO
- RFC 4592 (wildcard)
- RFC 4470 (minimal covering nsec) client/resolver side fine
- RFC 4956 (opt-in) TODO

hannesm:
some cname stuff is there, but `_build/default/app/odnssec.exe --host trac.ietf.org --type NS -vv --nameserver 9.9.9.9` results in "error from resolver invalid reply, couldn't find SOA" (using other resolvers works fine, but also results in different replies) -- it is the answer of the query for DS trac.ietf.org...

hannesm:
about the nsec denial of existence for *.se issue, https://lists.dns-oarc.net/pipermail/dns-operations/2022-January/021502.html should be read and the code adapted to accept the answers

hannesm:
now the tests are working, maybe need to re-review the code, merge it, and get back to the resolver implementation (+client)

hannesm:
ok, after a round of code review and implementing the tests from 5155 (nsec3):
- all tests are passing that have valid chains
- unsigned delegations are atm not accepted
- [ ] leftovers: what about nsec_chain when the nsec has been wildcard-expanded (and used_name is not owner_name)? figure out an example
- [ ] leftover: in nsec3 handling, there's the requirement (8.5/8.6) to sometimes check for certain bits set (Ns/Soa), this is not implemented
- [ ] leftover: nsec no-data should also ensure that cname bit is not set

the next question is about API:
- what is the common use case (only signed / allow unsigned replies)?
- should the dnssec API be disjoint (different module & package from the dns-client / dns-resolver)?
- should there be a dns-client implementation that uses dnssec?
- recursive dnssec resolver (aggressive caching / only store signed stuf / ...)

but I also think this PR is already large enough - and should be merged once the three tasks above are satisfied (and we convinced ourselves that the code validates correctly). //cc @reynir

hannesm:
TODO: research whether the rrsig signer_name should be verified somewhere?

hannesm:
rebased, squashed and merged manually.

PR #263, opened 2021-07-27
dns-cache: use a more specific type for get and set

hannesm:


reynir (approved):
Looks good to me (and I'm very pleased with the types)

PR #267, opened 2021-08-18
client: With_tls functor

hannesm:
This adds DNS-over-TLS (RFC 7858) to the dns-client. This compiles, but have not been tested (and clients need to be adapted, esp. the TLS authenticator needs to be provided).

### DIFF
--- a/client/dns_client.ml
+++ b/client/dns_client.ml
@@ -161,6 +161,140 @@ module type S = sig
   val lift : 'a -> 'a io
 end
 
+module With_tls (T : S) = struct
+  type +'a io = 'a T.io
+
+  type io_addr = Tls.Config.client * T.io_addr
+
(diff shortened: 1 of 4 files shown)


hannesm:
mostly a proof of concept here, but:
- odns works (with various public resolvers)

what is left to be done: provide command-line arguments for authenticator (trust anchors, public key (or cert) fingerprints, evenually hostname) -- eventually there should be a post-resolving authentication with dnssec, and dane (see https://datatracker.ietf.org/doc/html/rfc8310#section-6 for details)

hannesm:
with f6fabea the `odns` tool supports dns-over-tls, including command line arguments for the various authentication mechanisms.

hannesm:
This was an interesting prototype, but with #269 being merged we need to rework it.

PR #269, opened 2021-09-16
client updates: use a single TCP connection and multiplex over that

hannesm:


hannesm:
The CI failures are unrelated (AFAICT). This is already a good chunk in the right direction (with the goal being dns-over-tls resolution). Since all dependencies are released, I'll merge this and we can re-develop/continue on the main branch with dns-over-tls (and then cut a release).

hannesm:
Forgot to mention that this patch set was pair-programmed with @reynir, so we had two pairs of eyes looking at the code already.

Issue #276, opened 2021-11-09
TCP and EDNS


orbitz:
I could not get ocaml-dns working on my laptop which is using unbound and after some debugging I believe the issue is that in Tcp mode, ocaml-dns always attaches an edns record, which unbound does not support.  I flipped over to a UDP implementation and this was resolved.  This was an issue for me because whatever the AWS ECS instances run has the same issue.

1. Does what I have said make sense?
1. If it does, how do you feel about being able to thread a "disable edns" option through when creating a client?

hannesm:
that is a great suggestion, would https://github.com/mirage/ocaml-dns/pull/278/commits/4e3acd1936f7f5e1067202bfaea2f524ea454c1b suit your needs? (You can then pass `` ~edns:`None `` to `create`.

reynir:
Commit c717141b3a662e5eccb58565d698b5e8959edbbc introduced adding edns-tcp-keepalive. I saw that unbound introduced support for edns-tcp-keepalive in 1.8.0 in September 2018. What do you observe when it doesn't work? Can you try and revert commit c717141b3a662e5eccb58565d698b5e8959edbbc and see if that changes anything for you?

orbitz:
@hannesm This looks perfect, thank you!

orbitz:
For more information, here is the TCP Dump I did using dns-client.unix on the current latest release of dns-client.  The error is a "Format error" according to wireshark

https://www.dropbox.com/s/5e556fn5sqv9vwy/dns.log?dl=0

hannesm:
part of 6.1.0 release

PR #278, opened 2021-11-09
dns-client: make edns optional and configurable by API

hannesm:
val Dns_client.Make.create now includes
`` ?edns:[`None | `Auto | `Manual of Dns.Edns.t])``

as suggested by @orbitz in #276

### DIFF
--- a/app/dune
+++ b/app/dune
@@ -40,3 +40,10 @@
   (package      dns-cli)
   (libraries    dns dns-client.lwt dns-cli cmdliner mtime.clock.os
                 lwt.unix hex bos))
+
+(executable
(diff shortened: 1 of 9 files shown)


hannesm:
//cc @reynir the 2nd and 3rd commit should fix the behaviour we observed yesterday (TLS authentication failure). in the end, I think it is fine to ahve a map IP -> Domain_name in the dns-client source code since I can't see a different way how to proceed.

orbitz:
Woo, thank you for the quick turnaround

hannesm:
merged manually with review by @reynir

PR #283, opened 2021-12-17
Upgrade to tcpip.7.0.0

dinosaure:
`sed s/Mirage_stack/Tcpip.Stack/gc`
`sed s/mirage-stack/tcpip/gc`

/cc @hannesm


hannesm:
thanks! merged and released as 6.1.3

PR #285, opened 2022-01-04
dns-client (mirage/lwt): avoid connecting to resolver race

hannesm:
When multiple DNS requests are issued, each of them attempts a connection to
the resolver (since the connection initialization is not tracked), leading to
multiple file descriptors being used and the fd/flow being set to one of the
connections. Instead, once a connection is attempted, establish a Lwt_condition
and wait in other promises until that is resolved before proceeding.

### DIFF
--- a/lwt/client/dns_client_lwt.ml
+++ b/lwt/client/dns_client_lwt.ml
@@ -18,6 +18,7 @@ module Transport : Dns_client.S
     timeout_ns : int64 ;
     (* TODO: avoid race, use a mvar instead of condition *)
     mutable fd : [ `Plain of Lwt_unix.file_descr | `Tls of Tls_lwt.Unix.t ] option ;
+    mutable connected_condition : unit Lwt_condition.t option ;
     mutable requests : (Cstruct.t * (Cstruct.t, [ `Msg of string ]) result Lwt_condition.t) IM.t ;
     mutable he : Happy_eyeballs.t ;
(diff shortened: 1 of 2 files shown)


lwt/client/dns_client_lwt.ml
```diff
@@ -266,6 +273,8 @@ module Transport : Dns_client.S
       Lwt.async (fun () -> Lwt_list.iter_p (handle_action t) actions);
       waiter >>= function
       | Error `Msg msg ->
+        Lwt_condition.broadcast connected_condition ();
```
reynir, line 276: If an error happens while multiple 'threads' are trying to connect it would result in retrying immediately one by one. Ideally, there would be some back-off semantics. Perhaps the condition could propagate a value of type `` [`Connected | `Retry of timeout]``, and in the case of error use Lwt_condition.signal and reuse the condition.

reynir (approved):
I like the motivation for this patch and I like the implementation. I have a concern about the failure mode, but it is no worse than the status quo.

lwt/client/dns_client_lwt.ml
hannesm: thanks for your review. My thinking is (taking non-preemptive scheduling of Lwt into account):
- we have a connection failure, and broadcast to all waiting DNS resolutions
- one of which is scheduled, which calls `connect_via_tcp_to_ns` again, which does the match on fd and connected_condition (hint: both are None)
- the branch where a new condition is allocated and t.connected_condition is mutated is taken, which leads to another connection attempt

all other tasks that are scheduled now will "wait" for yet another broadcast.

PR #290, opened 2022-01-18
resolver: IPv6 support and DNSSec

hannesm:


hannesm:
- [ ] DS records should be requested at delegation point (not auth servers)
- [ ] DS records caching (if ttl < NS ttl, they'll never be re-requested -- and zone is treat as no dnssec)
- [ ] AD bit in reply

later:
- "aggressive caching" of negative queries
- CD bit in request
- DS for root zone may time out (after 68 years), there should be a path to update without code update (same for NS --> maybe use an (optional) zone file and parse that?)

resolver/dns_resolver_cache.ml
```diff
@@ -11,6 +11,47 @@ let _pp_err ppf = function
...
+        false
+  in
+  up name
+
+let nsec3_covering t ts name =
+  let rec up name =
+    match snd (Dns_cache.get_nsec3 t ts name) with
+    | Ok nsec3 ->
+      let Nsec3.{ iterations ; salt ; _ } = snd (List.hd nsec3) in
+      let soa_name = Domain_name.drop_label_exn ~rev:true name in
```
reynir, line 32: Do we really want `~rev:true`? doesn't that turn foo.example.org into foo.example?

hannesm: you're right, I'm not sure what I had in mind here.

resolver/dns_resolver_cache.ml
```diff
@@ -278,9 +278,18 @@ let handle_query t ~rng ip_proto ts (qname, qtype) =
     `Reply (flags, data), t
   | `Query name, t ->
     Logs.info (fun m -> m "query for %a (%a): %a" Domain_name.pp qname Packet.Question.pp_qtype qtype Domain_name.pp name);
-    let zone, name', types, ip, t = resolve t ~rng ip_proto ts name qtype in
+    (* DS should be requested at the parent *)
+    let name', recover =
+      if Domain_name.count_labels name > 1 && qtype = `K (Rr_map.K Ds) then
+        let n' = Domain_name.drop_label_exn name in
```
reynir, line 284: I'm not sure about this. The parent zone is not necessarily the domain without one label, e.g. if example.org delegates bar.foo.example.org then the parent of bar.foo.example.org is not foo.example.org. Maybe I am missing something; I will have to read the context.

hannesm: True, the parent is not necessarily the zone cut, but this does not matter since DS records are only at the zone cuts. Take `example.com` and `cut.example.com` as zone cuts, then:
- query for DS of example.com should be done at the com nameservers
- query for DS of cut.example.com should be done at the example.com nameservers
- query for DS of foo.example.com does not exist (and is now done at the example.com nameservers)
- query for DS of bar.cut.example.com does not exist (but is done at the cut.example.com nameservers)

hannesm:
I squashed and force-pushed. When CI is happy, fine to merge. We can think of a strategy for ANY at a later point :)

reynir (approved):

Issue #243, opened 2020-11-04
DNS client improvements (UDP & resolv.conf parsing & multiple name servers)


hannesm:
This is copied from the discussion with @cfcs in https://github.com/mirage/ocaml-dns/pull/241 and scheduled for the next major release

To iteratively move forward, my suggestion is (where the checkboxes are intended for the next release, `(later)` for some future release):
- [ ] together with #239 we're breaking the API anyways, let's have `create` receive a nameserver list
- [ ] work on enhancing the resolv-conf parser (timeout, ..) and the DNS client semantics (remember which NS was tried last, ...)
- [ ] get UDP support in the client(s) -- we can retain TCP as default (eventually with caching the TCP connection, as done in dns-stub), but mainly need retransmission (on lookup timeout)
- [ ] `domain` / `search` support for resolv.conf
- [ ] keep a hash of the `resolv.conf` contents, and on lookup failure re-read that file (to support laptops who joined a different network, and resolv.conf was updated by DHCP etc.)
- (later) add DNSSec validation (needed for a reasonable recursive resolver anyways - need some magic to have it disabled when in a network where you've to browse somewhere and click ok)
- (later) add DNS-over-TLS support (here question arise in the form of "where does the certificate come from" etc.)

hannesm:
With #269 there are some fixes: now all nameservers from /etc/resolv.conf are used. Also, the lwt DNS client uses UDP and TCP (the unix one as well, but only the first nameserver).

I don't think that with DNS-over-TLS on the horizon we should strive for UDP support in the MirageOS DNS client. Instead I'd be in favour to allow TLS and TCP only (and maybe remove UDP support from the Lwt_unix and Unix clients as well). Reasoning is less code and more reliability.

hannesm:
Closing this issue, as most tasks have been accomplished. The remaining "support for search in /etc/resolv.conf" can be done once someone reports the demand for it.

PR #291, opened 2022-02-03
dns-client: if /etc/resolv.conf modifies, update the internal list of resolvers

hannesm:
Only affects lwt and unix.
Previously, when you started an application, /etc/resolv.conf was read once at
startup. When you switch to a different network, and /etc/resolv.conf is updated
by DHCP, this is not reflected in the dns-client.

Now, the digest of /etc/resolv.conf is stored internally, and if it changes, it
is parsed again. The lwt implementation needed to be modified slightly since
connect_to_ns_via_tcp carried a nameserver list, and a different task may have
been woken up. Now, there are two functions and it should play out nicely.

//cc @reynir addresses #243

### DIFF
--- a/lwt/client/dns_client_lwt.ml
+++ b/lwt/client/dns_client_lwt.ml
@@ -14,7 +14,8 @@ module Transport : Dns_client.S
   type +'a io = 'a Lwt.t
   type stack = unit
   type t = {
-    nameservers : io_addr list ;
+    mutable nameservers : io_addr list ;
+    mutable resolv_conf : Digest.t option ;
     timeout_ns : int64 ;
     (* TODO: avoid race, use a mvar instead of condition *)
     mutable fd : [ `Plain of Lwt_unix.file_descr | `Tls of Tls_lwt.Unix.t ] option ;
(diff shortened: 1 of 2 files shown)


lwt/client/dns_client_lwt.ml
```diff
@@ -103,38 +104,97 @@ module Transport : Dns_client.S
...
+        t.resolv_conf <- None;
+        t.nameservers <- default_resolver ()
+
   let create ?nameservers ~timeout () =
-    let nameservers =
+    let nameservers, resolv_conf =
       match nameservers with
       | Some (`Udp, _) -> invalid_arg "UDP is not supported"
-      | Some (`Tcp, ns) -> ns
+      | Some (`Tcp, ns) -> ns, None
```
reynir, line 189: Perhaps we should preserve whether the nameservers were passed explicitly and then not try to read resolv.conf.

reynir (commented):
This looks good to me except I think we should not update the list of nameservers if the caller of `Dns_client_{lwt,unix}.create` explicitly passed a list of nameservers.

haesbaert:
Wouldn't it make more sense to just look at mtime on /etc/resolv.conf instead of constantly calculating digests ? 
Or did I miss something ?

reynir:
> Wouldn't it make more sense to just look at mtime on /etc/resolv.conf instead of constantly calculating digests ? Or did I miss something ?

@haesbaert we discussed this on IRC and most of it is summarized in #296. Note that the lwt implementation will reuse the connection if possible and "only" re-read resolv.conf when reconnecting.

haesbaert:
> > Wouldn't it make more sense to just look at mtime on /etc/resolv.conf instead of constantly calculating digests ? Or did I miss something ?
> 
> @haesbaert we discussed this on IRC and most of it is summarized in #296. Note that the lwt implementation will reuse the connection if possible and "only" re-read resolv.conf when reconnecting.

Ack, sorry for the noise :D

reynir:
I made a pull request to this pull request https://github.com/hannesm/ocaml-dns/pull/1 where I implement the behavior I mentioned earlier.

PR #297, opened 2022-02-10
Add some utilities to be able to provide a proper DNS device for MirageOS

dinosaure:
This PR wants to provide a `connect_device` which can be used by a MirageOS device and expects raw arguments (due to the fact that it's not possible to extend _converters_ in Mirage) and try to parse them. If these arguments are wrong, we just fails. The semantic of arguments is not yet clear but it's a draft. The idea is to have 3 arguments:
- `--dns-authenticator` which can add an authenticator to a specific IP address
- `--dns-nameserver` which can add a possible nameserver to our DNS object. If we found an authenticator associated to the given IP address, we use it and consider that the flow must use TLS. Otherwise, it's a _plaintext_ communication
- `--dns-protocol` which defines which transmission protocol we want to use (default to `udp`)

With all of these arguments, the `Dns_client_mirage.S`, the new `connect_device` function and mirage/mirage#1292, we are able to provide a simple DNS device which can be re-used then by `Happy_eyeballs`. Related to mirage/mirage#1291.


dinosaure:
I said on others PRs:
> NOTE: by design, connect_device should not appear into the signature but only into the functor. I put it but if you are more confident to put it outside (as any devices in mirage), I can do it.

mirage/client/dns_client_mirage.ml
```diff
@@ -1,12 +1,105 @@
...
+      let certs = List.filter_map Result.to_option certs in
+      ipaddr, X509.Authenticator.chain_of_trust ~time certs
+    | [ nameserver; "none" ] ->
+      let ipaddr = Ipaddr.of_string_exn nameserver in
+      ipaddr, none
+    | _ -> Fmt.failwith "Invalid TLS authenticator: %S" str
+
+  let authenticators_of_strings lst =
+    let lst = List.map authenticator_of_string lst in
+    List.fold_left (fun a (k, v) -> Ipaddr.Map.add k v a) Ipaddr.Map.empty lst
```
hannesm, line 95: thanks, but that code should be in x509, not in dns.

mirage/client/dns_client_mirage.ml
```diff
@@ -310,4 +403,13 @@ module Make (R : Mirage_random.S) (T : Mirage_time.S) (M : Mirage_clock.MCLOCK)
...
   include Dns_client.Make(Transport)
+
+  let connect_device ?(protocol= "udp") ?(authenticators= []) ?(nameservers= []) ?(timeout= 5_000_000L) stack =
+    let protocol = match String.lowercase_ascii protocol with
+      | "udp" -> `Udp
+      | "tcp" -> `Tcp
+      | _ -> Fmt.failwith "Invalid protocol: %S" protocol in
+    let authenticators = authenticators_of_strings authenticators in
+    let nameservers = List.map (nameserver_of_string authenticators) nameservers in
+    Lwt.return (create ~nameservers:(protocol, nameservers) ~timeout stack)
```
hannesm, line 414: taking a look at the implementation, the `protocol` should defalut to `tcp` (since udp is not supported in this implementation).

mirage/client/dune
```diff
@@ -1,5 +1,5 @@
 (library
   (name        dns_client_mirage)
   (public_name dns-client.mirage)
-  (libraries   domain-name ipaddr mirage-random mirage-time tcpip mirage-clock dns-client happy-eyeballs tls-mirage ca-certs-nss)
+  (libraries   astring hex domain-name ipaddr mirage-random mirage-time tcpip mirage-clock dns-client happy-eyeballs tls-mirage ca-certs-nss)
```
hannesm, line 4: there was previous some care to not depend on astring and hex, I'm not convinced we should re-add these dependencies here.

mirage/client/dns_client_mirage.ml
```diff
@@ -1,12 +1,105 @@
...
+  match String.split_on_char ':' str with
+  | [ addr; port ] ->
+    ( try Ipaddr.of_string_exn addr, int_of_string port
+      with _ -> Fmt.failwith "Invalid inet address: %S" str )
+  | [ addr ] ->
+    ( match Ipaddr.of_string addr with
+    | Ok addr -> addr, 53
+    | Error (`Msg err) -> failwith err )
+  | [ ] -> Fmt.failwith "Empty inet address"
+  | _ -> Fmt.failwith "Invalid inet address: %S" str
```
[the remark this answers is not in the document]
dinosaure: What do you suggest about that? I think about:
```ocaml
let lst = String.split_on_char ':' str in
( match List.rev lst with
| port :: ipaddr ->
  ( match Ipaddr.of_string (String.concat ":" (List.rev ipaddr),
          Ipaddr.of_string (String.concat ":" lst),
          int_of_string_opt port with
  | _, Ok ipaddr, _ -> ipaddr, 53
  | Ok ipaddr, Error _, Some port -> ipaddr, port
  | _, Error _, None -> Fmt.invalid_arg "Invalid inet-addr: %S" str )
| [ ipv4 ] -> Ipaddr.of_string ipv4, 53
| [ ] -> Fmt.invalid_arg "Invalid empty inet-addr" )
```

mirage/client/dns_client_mirage.ml
dinosaure: If you agree with the format (already used by `ocaml-git` for the HTTPS authenticator), I can propose it on `x509` yes.

mirage/client/dns_client_mirage.ml
dinosaure: > (since udp is not supported in this implementation)

I was not aware about that, I believed that both are supported 😕 .

mirage/client/dune
dinosaure: `astring` & `hex` was used for `authenticator`, if I move `authenticator_of_string` into `x509`, it should not be a problem then 🙂

dns-mirage.opam
```diff
@@ -38,3 +38,6 @@ of the hash algorithm is used).
 Please read [the blog article](https://hannes.nqsb.io/Posts/DNS) for a more
 detailed overview.
 """
+pin-depends: [
+  [ "x509.dev" "git+https://github.com/dinosaure/ocaml-x509.git#4b96bb6a76613f512220aba5d93d115d946eeb38" ]
+]
```
hannesm, line 43: would you mind to use `x509 >= 0.16.0` above instead? thanks.

mirage/client/dns_client_mirage.ml
```diff
@@ -1,12 +1,69 @@
...
+  | [ ipv4 ] ->
+    let+ ipaddr = Ipaddr.of_string ipv4 in Ok (ipaddr, 53)
+  | port :: ipaddr ->
+    ( match Ipaddr.of_string (String.concat ":" (List.rev ipaddr)),
+            Ipaddr.of_string (String.concat ":" lst),
+            int_of_string_opt port with
+    | _, Ok ipaddr, _ -> Ok (ipaddr, 53)
+    | Ok ipaddr, Error _, Some port -> Ok (ipaddr, port)
+    | _, Error _, _ -> Error (`Msg (Fmt.str "Invalid inet-addr: %S" str)) )
+  | [ ] -> Error (`Msg "Invalid empty inet-addr") )
```
hannesm, line 18: Sorry for being picky, but to reuse this functionality, I'd prefer to have it in the ocaml-ipaddr package. Maybe a `with_port_of_string : ?default:int -> string -> (t * port, [> `Msg of string ]) result` in all three modules? :)

mirage/client/dns_client_mirage.ml
```diff
@@ -310,4 +367,19 @@ module Make (R : Mirage_random.S) (T : Mirage_time.S) (M : Mirage_clock.MCLOCK)
   end
 
   include Dns_client.Make(Transport)
+
+  let connect_device ?(protocol= "udp") ?(authenticators= []) ?(nameservers= []) ?(timeout= 5_000_000L) stack =
```
hannesm, line 371: - UDP shouldn;t be the default
- please use Duration instead of 5_000_000L
- the interface is rather brittle, since it allows multiple nameservers and authenticators, but they are unrelated. also they need to use the same protocol.

What about `connect_device : <> -> ?timeout -> stack -> t` where the `<>` is an (optional) list of:
- protocol (udp, tcp, tls)
- ip address
- port (optional)
- authenticator (if tls was used)

The authenticator may as well need to allow to pass the hostname to validate against (instead of using the IP address).

mirage/client/dns_client_mirage.ml
dinosaure: > in all three modules?

Which module you want? You would like to integrate that into `ipaddr` (like `Ipaddr.with_port_of_string`)?

hannesm: Yes, that would be great imho.

mirage/client/dns_client_mirage.ml
dinosaure: I don't really understand the format and I just would like to clarify a point: we are restricted by what `Mirage.Key.Arg` can give to us (because this `connect_device` will be used then by `functoria`) and we don't many possibilities (only `string`, `ipaddr` or `int`).

So can you elaborate what you expect about:
> What about connect_device : <> -> ?timeout -> stack -> t where the <> is an (optional) list of:

hannesm: The issue I see with the current signature: `?protocol -> ?authenticators -> ?nameservers` is that there is a single protocol that all nameservers use, and also there's only loosely coupled "authenticators" with "nameservers".

I'd instead be in favour of having a single `string` that uses can pass that includes all information about a resolver:
- `PROTO:IP:PORT:AUTH`

So, as an example:
- `141.1.1.1` (default protocol, default port)
- `UDP:141.1.1.1` (UDP, default port)
- `TLS:8.8.8.8:key-fp:aaaaa` (TLS, default port, pinned to a specific public key)

In this way, a user can provide multiple name servers with distinct protocols and tightly coupled authenticators.

As noted earlier, and again above, the `protocol` (in respect to the dns-client dns-client-mirage interaction) is always `` `Tcp `` (since udp is not supported, and tls uses the same format as tcp).

dinosaure: Ok I see, I will come back with something better.

hannesm: Thanks @dinosaure. This PR is highly appreciated (to finally use the dns-client directly in MirageOS unikernels), and pretty close for being merged and released. Sorry for my recent lack of time on reviewing this PR (and thanks for taking the time to push the changes to x509/ipaddr/...) :D

dinosaure:
I force-pushed a new version according the new version of `x509.0.16.0` and the new release of `ipaddr.5.3.0` (which is not yet released but it will be soon). I integrated your comment @hannesm and provide a better way to _unserialize_ nameservers given by the `mirage` tool. Tell me what you think about that.

hannesm:
Tanks again for your work, I reviewed it and pushed two commits. The behavioural change is that `udp` is no more supported by nameserver_of_string -- since "let create" errors on `` `Udp ``, I don't think it is very useful to support udp in the nameserver_of_string. Also, "connect_device" has been renamed to "connect". The default authenticator is now Ca_certs_nss.authenticator.

WDYT?

dinosaure:
I don't have strong opinion with what you did, it's perfect for me to, then, be able to propose a new device on the `mirage` side 👍 .

mirage/client/dns_client_mirage.ml
```diff
@@ -5,11 +5,53 @@ module Log = (val Logs.src_log src : Logs.LOG)
...
   module CA = Ca_certs_nss.Make(P)
 
+  let nameserver_of_string str =
+    let ( let* ) = Result.bind in
+    match String.split_on_char ':' str with
+    | "tls" :: rest ->
+      let str = String.concat ":" rest in
+      ( match String.split_on_char '!' str with
+      | [ nameserver ] ->
+        let* ipaddr, port = Ipaddr.with_port_of_string ~default:853 nameserver in
```
reynir, line 34: Does this support ipv6? And how does it work with ipv6's `:` separator and port separator?

reynir (commented):
This looks good to me. I have two remarks regarding `nameserver_of_string`.

mirage/client/dns_client_mirage.ml
dinosaure: You should check my recent PR about that: https://github.com/mirage/ocaml-ipaddr/pull/108
IF you want to specify a port with an IPv6 address, you must _expand_ the IPv6 address. For instance,
- `::1:8080` returns an IPv6 with the default port
- `0:0:0:0:0:0:0:1:8080` returns `::1` with the port `8080`

mirage/client/dns_client_mirage.ml
```diff
@@ -5,11 +5,53 @@ module Log = (val Logs.src_log src : Logs.LOG)
...
+        let time () = Some (Ptime.v (P.now_d_ps ())) in
+        let authenticator = authenticator time in
+        let tls = Tls.Config.client ~authenticator () in
+        Ok (`Tls (tls, ipaddr, port))
+      | [] -> assert false )
+    | "tcp" :: nameserver ->
+      let str = String.concat ":" nameserver in
+      let* ipaddr, port = Ipaddr.with_port_of_string ~default:53 str in
+      Ok (`Plaintext (ipaddr, port))
+    | _ ->
```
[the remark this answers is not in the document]
hannesm: I agree, done in e12ad9b

hannesm:
(force-pushed after merge of #300 to allow CI to succeed)

PR #309, opened 2022-04-12
server: demote log level, make Dns_one.add_additional_glue more correct

hannesm:


hannesm:
best viewed commit by commit. includes test cases and bugfixes in the aftermath of #308

reynir (approved):
To me this looks reasonable and good

PR #310, opened 2022-04-26
Add the LOC record

RyanGibb:
See [rfc1876](https://datatracker.ietf.org/doc/html/rfc1876)


src/dns.ml
```diff
@@ -1303,6 +1303,113 @@ module Nsec3 = struct
...
+    let retval =
+      (Int.shift_left 1 31) + (
+        (((Int32.to_int deg * 60) + Int32.to_int min) * 60) * 1000 +
+        Int.of_float (sec *. 1000.)
+      ) * if dir then 1 else -1
+    in
+    Int32.of_int retval
+
+  let alt_parse alt =
+    let _ = Printf.printf "%d\n" (Int.of_float (10000000. +. alt *. 100.)) in
```
avsm, line 1327: is this a debug statement left over?

src/dns.mli
```diff
@@ -791,6 +791,35 @@ module Edns : sig
       buffer. *)
 end
 
+(** Loc
+
+    A locator record (LOC) is used to express location information associated with domain. *)
+module Loc : sig
+  type t = {
+    lat : int32;
+    long : int32;
```
avsm, line 800: can lat/long simply be `int` to avoid boxing the value? OCaml guarantees at least 31 bits for an `int`

avsm (commented):
Have had a first read through, and the overall approach looks good to me. Thanks for the PR @RyanGibb!  Need to test it out and look through the RFC before a full approval. Some more eyes on this welcome @mirage/core

src/dns.ml
RyanGibb: Yep, I missed this one!

src/dns.mli
RyanGibb: Using `int` doesn't seem to deal with negatives properly, I think because 32nd bit is used to encode the direction of the lat/long.

Using an `int` version of `lat_long_parse` and `lat_long_print`:
```
let lat_long_parse_int_version ((deg, min, sec), dir) =
    (Int.shift_left 1 31) + (
      (((Int32.to_int deg * 60) + Int32.to_int min) * 60) * 1000 +
      Int.of_float (sec *. 1000.)
    ) * if dir then 1 else -1

let lat_long_print_int_version lat_long =
    let dir = lat_long < 0 in
    let lat_long = Int.abs lat_long in
    let lat_long = (Int.shift_left 1 31) - lat_long in
    let sec = Float.of_int (lat_long mod (60 * 1000)) /. 1000. in
    let min = (lat_long / (1000 * 60)) mod 60 in
    let deg = (lat_long / (1000 * 60 * 60)) mod 60 in
    deg, min, sec, dir
```

With 0 0 1 N latitude we get:
```
# lat_long_parse_int_version ((0l, 0l, 1.), true);;
- : int = 2147484648
```

And as the directionality is encoded by the sign (e.g. the direction is encoded in the first bit) this messes up our decoding:
```
# lat_long_print_int_version 2147484648;;
- : int * int * float * bool = (0, 0, -1., false)
```

We should never have a negative deg/min/sec (as specified in [RFC1876 section 3](https://datatracker.ietf.org/doc/html/rfc1876#section-3)).

In contrast, Int32 deals with the negative properly:
```
# lat_long_parse ((0l, 0l, 1.), true);;
- : int32 = -2147482648l

# lat_long_print (-2147482648l);;
- : int * int * float * bool = (0, 0, 1., true)
```

Note that converting the `int` to an `int32` gives the desired negative value, which is what we're doing at the end of `lat_long_parse_int_version`:
```
Int32.of_int 2147484648;;
- : int32 = -2147482648l
```

RyanGibb: maybe it would be better to explicitly use the `Int32` arithmetic operations?

hannesm:
Thanks for your PR, would you mind to add some tests into `tests/test.ml`covering both successful packet decoding and unsuccessful packet decoding? Also for encoding the packet... Take a brief look at that test module, it should be pretty straightforward to add that.

About the zone file parsing, there aren't many tests yet -- maybe add some to server.ml.

RyanGibb:
Yes, sure!

src/dns.mli
RyanGibb: Looking at this with a fresh eye, there was a bug in my `lat_long_print` decoding that replied on the leftmost bit being negative, but actually this should only be the case after 2^32 is subtracted from the value.

TL;DR: using `int` is possible and desirable!

RyanGibb:
Hi @hannesm, that's tests covering successful packet decoding, unsuccessful packet decoding, and packet encoding added.

RyanGibb:
And a simple parsing test.

RyanGibb:
Currently the altitude maximum tests are failing. This is due to `Int32` taking a positive number > 2^31 as a signed number and

RyanGibb:
I've solved the encoded altitude > 2**31 with a horrible hack that uses an `int64` instead and writes to the buffer at an offset of 4 bytes with
```
Cstruct.BE.set_uint64 buf (off + 8) loc.alt;
```
As opposed to
```
Cstruct.BE.set_uint32 buf (off + 12) loc.alt;
```
Which relies on the later statement
````
Cstruct.BE.set_uint32 buf (off + 8) loc.long;
```
Correctly setting the 4 bytes at offset 8 (which will always be 0 from loc.alt as it's at most 32 bits).

This requires some clearing and resetting these bits when reading the alt value:
```
    (* Horrible hack to deal with encoded altitude > 2**31,
       as Int32.to_float converts anything > 2**31 to -2**31 *)
    let alt = Cstruct.BE.set_uint32 buf (off + 8) 0l;
      Cstruct.BE.get_uint64 buf (off + 8) ;
    in
    Cstruct.BE.set_uint32 buf (off + 8) long;
```

RyanGibb:
Finally, fixed arm32 and x86_32, by removing intermediary cast to `int` between `int64` and `float` in `alt_print`.

src/dns.mli
```diff
@@ -791,6 +791,35 @@ module Edns : sig
...
+    horiz_pre : int;
+    vert_pre : int;
+  }
+  (** The type of a Loc record. *)
+
+  val parse : ((int32 * int32 * float) * bool)
+    -> ((int32 * int32 * float) * bool)
+    -> float
+    -> (float * float * float)
+    -> t
```
hannesm, line 812: I've no clue what the arguments for this function are. Lots of floating point values.

hannesm:
I just re-read Appendix A of RFC 1876 - and this is all about string to binary conversion, but avoids any floating point values. So I suggest to have a "decode/parse" function taking a string and returning the int32 value (and the precision, or what is needed). Does this make sense?

The whole DNS library is so far free of floating point values, and I appreciate to keep it that way.

RyanGibb:
> > I've solved the encoded altitude > 2**31 with a horrible hack that uses an int64 instead and writes to the buffer at an offset of 4 bytes with
> > `Cstruct.BE.set_uint64 buf (off + 8) loc.alt;`
> 
> I'm not sure I understand why this is necessary. Doesn't an `int32` work in terms of encoding and decoing -- but the interpretation should be adjusted (e.g. in the altitude)?

Yes, you're correct. The parsing and printing is where the issue arises. I've added a change that uses [ocaml-integers](https://github.com/ocamllabs/ocaml-integers) to parse and print the variable. This is also a bit hacky, as there is no `Unsigned.UInt32.to_float`. Instead, I've added a chain of casting from `float` to `Int64` to `Unsigned.UInt32` to `int32` for parsing, and vice versa for printing. This is in order to avoid the problematic `Int32.of_float` and `Int32.to_float`. Do let my know if there's a better way to do this.

> Why should all this deal with `float` -- AFAICT the RFC uses various integers (i.e. only represents a few coordinates with a given precision). From my perspective, an API that takes only integer values (int32 etc.) would be sufficient -- certainly some "geo-coordinate library" could deal with float conversions from geocoordinates to DNS RRs.

The floats arise from the parsing. Unlike the RFC this libraries parsing and encoding (/decoding and printing) are separated, which allows us to be more flexible. It would certainly be possible to adapt the parsing to read fixed precision integers, but I'm not sure what the benefit of this additional complexity would be. What is your reason for wanting to avoid floats? Is it performance? If so, is it sufficient that the `Loc` module only deals with floats while parsing and printing, and the encoding/decoding is all using `int32`s/`int`s?

> Another comment is the "Loc set" -- the RFC is not very specific here, but does it make sense for any domain name to have multiple LOC records? If the answer is no, let's (similar to SOA and CNAME) use only a single Loc.t (not a set). This makes the handling and API also much simpler.

There is not a clear use case for this record.  The RFC discusses a number of possibilities, but it's quite open ended. For this reason I think providing the greatest flexibility would be best. I can certainly imagine a scenario where someone would want to associate multiple LOC records with one domain. Also, other DNS server implementations support multiple LOC records for one domain:
```
$ dig loc gibbr.org +short
51 53 23.990 N 1 28 34.190 E 0.00m 10m 1m 1m
52 12 40.000 N 0 5 31.000 W 22.00m 10m 10m 10m
```

hannesm:
> Indeed, you are correct. The parsing and printing is where the issue arises. I've added a change that uses [ocaml-integers](https://github.com/ocamllabs/ocaml-integers) to parse and print the variable.

Great, but sorry to tell you: I do not like to add dependencies to `dns` . Could you retrieve the required bits from the integers library and embed them into the `Loc` module?

> The floats arise from the parsing and printing. It would certainly be possible to adapt these to read fixed precision integers, but I'm not sure what the benefit of this additional complexity would be. Can I ask what your reason for wanting to avoid floats is?

Floats are at some points problematic: being imprecise, not being used on the wire, various issues if you're using them in kernel mode (as running as MirageOS unikernel eludes to). But I see that your `parse` function is the only one using floats, so for LOC records transmitted on the wire or in a zone file that code wouldn't be exercised - please correct me if I'm wrong.

> I'm not sure I completely understand the idea behind the geocoordinate library; the representation here is quite closely coupled to the LOC record.

I was imaging that a potential use would be that there's a GPS receiver somewhere used to fill in the geocoordinates -- or someone looking up on a map the geocoordinate and wanting to insert them as floating point number.

> And I can certainly imagine a scenario where someone would want to associate multiple LOC records with one domain. Also, other DNS server implementations (notably bind) support multiple LOC records for one domain:

Ok, fine with me. I was only wondering what the interpretation may be.

RyanGibb:
> Great, but sorry to tell you: I do not like to add dependencies to `dns` . Could you retrieve the required bits from the integers library and embed them into the `Loc` module?

No problem, can do! This required pulling a fair bit of C in, so instead I just manually converted the unsigned int32 to an int64. The `Int64.to_int32` statement is fine, but the `Int64.of_int32` is where the trouble arises.

Thanks for the help with this, and apologies for the back and forth. This is my first time writing OCaml so I'm still getting to grips with the conventions.

> Floats are at some points problematic: being imprecise, not being used on the wire, various issues if you're using them in kernel mode (as running as MirageOS unikernel eludes to). But I see that your `parse` function is the only one using floats, so for LOC records transmitted on the wire or in a zone file that code wouldn't be exercised - please correct me if I'm wrong.

The floats are used for parsing and printing. For LOC records transmitted on the wire the code wouldn't be exercise. It would be exercised when reading the record from the file, or printing the record. Is this ok?

The printing would be relatively easy to change, the parsing a bit trickier.

> > And I can certainly imagine a scenario where someone would want to associate multiple LOC records with one domain. Also, other DNS server implementations (notably bind) support multiple LOC records for one domain:
> 
> Ok, fine with me. I was only wondering what the interpretation may be.

Cool :-) I guess one interpretation could be a multihomed network. At a higher level, maybe a home and work address.

RyanGibb:
By the way, is there a code formatter for the project?

avsm:
@hannesm:
> Great, but sorry to tell you: I do not like to add dependencies to dns 

Usually a good rule of thumb, but `ocaml-integers` is a pretty foundational dependency (e.g. ctypes also uses it). It's quite fiddly getting that handling correct, and I'm not in favour of duplicating that code.  There must be plenty of other places outside of Loc where we could potentially use signed/unsigned parsing to OCaml integers.

(I've wanted a ctypes<->cstruct bridge for years now, since ctypes interprets different integer/floats very well, and cstruct parses them from bytes very efficiently)

hannesm:
@avsm
> Usually a good rule of thumb, but ocaml-integers is a pretty foundational dependency (e.g. ctypes also uses it). It's quite fiddly getting that handling correct, and I'm not in favour of duplicating that code. There must be plenty of other places outside of Loc where we could potentially use signed/unsigned parsing to OCaml integers.

And does ocaml-integers work in a MirageOS 3 and MirageOS 4 environment? I figured it uses C code, and am not able to find any (MirageOS 3 required) cross-compilation runes.

avsm:
It does use a small C binding. But at some point we have to stop supporting Mirage3 and just move our library base forward (the older releases of DNS will continue to work with the older Mirage3 of course).   Perhaps this is a general point of discussion at the next Mirage community meeting to see how long we need to support Mirage3 moving forwards.

There is also the question of jsoo as well — it looks like integers is supported on that via https://ocaml.org/p/integers_stubs_js/1.0

RyanGibb:
> and I'm not in favour of duplicating that code

In this specific case, the duplication is relatively light:
```
    (* convert a uint32 alt to an int64 *)
    let alt = if alt < 0l then
        Int64.of_int32 alt + Int64.shift_left 1L 32
      else Int64.of_int32 alt
    in
```

Is there anything else I can improve to get an approval @hannesm ?

hannesm:
Dear @RyanGibb, I guess all you need is a bit patience. I spent some hours in the review of this PR, and since haven't found time to review your most current changes.

RyanGibb:
No problem, and no rush. I'll wait to hear back from you.

RyanGibb:
Hi @hannesm, apologies to bother you, but I thought I'd give this a gentle bump. It would be great if you would have time to review the most recent changes. However, if you don't have the time, I completely understand.

Thanks for all your help this far.

hannesm:
Thanks @RyanGibb, I'll take a look later this week. I am just returning from vacation. Thanks for your additional work on it :)

.envrc
```diff
@@ -0,0 +1 @@
+use_nix
```
hannesm, line 1: what is the purpose of this file? can we please drop it from the git repo?

main.tex.blg
```diff
@@ -0,0 +1,5 @@
+[0] Config.pm:311> INFO - This is Biber 2.16
+[0] Config.pm:314> INFO - Logfile is 'main.tex.blg'
+[43] biber:340> INFO - === Thu Jun  2, 2022, 13:23:01
+[97] Utils.pm:411> ERROR - Cannot find 'main.tex.bcf'!
+[97] Biber.pm:132> INFO - ERRORS: 1
```
hannesm, line 5: all these `main.*` should not be part of this PR.

shell.nix
```diff
@@ -0,0 +1,13 @@
...
+    ocaml
+    opam
+    dune_2
+    ocamlPackages.utop
+    pkg-config
+    gcc
+    bintools-unwrapped
+    gmp
+  ];
+}
```
hannesm, line 13: I've no idea why this file is needed in the repository.

src/dns.ml
```diff
@@ -2203,7 +2368,7 @@ module Rr_map = struct
   let to_int : type a. a key -> int = function
     | A -> 1 | Ns -> 2 | Cname -> 5 | Soa -> 6 | Ptr -> 12 | Mx -> 15
     | Txt -> 16 | Aaaa -> 28 | Srv -> 33 | Ds -> 43 | Sshfp -> 44
-    | Rrsig -> 46 | Nsec -> 47 | Dnskey -> 48 | Nsec3 -> 50
+    | Rrsig -> 46 | Nsec -> 47 | Dnskey -> 48 | Nsec3 -> 50 | Loc -> 29
```
hannesm, line 2371: could we move that to a place so that the RHS are ordered by integers? i.e. a line above in the middle

src/dns.ml
```diff
@@ -2214,7 +2379,8 @@ module Rr_map = struct
     | 12 -> Ok (K Ptr) | 15 -> Ok (K Mx) | 16 -> Ok (K Txt) | 28 -> Ok (K Aaaa)
     | 33 -> Ok (K Srv) | 43 -> Ok (K Ds) | 44 -> Ok (K Sshfp)
     | 46 -> Ok (K Rrsig) | 47 -> Ok (K Nsec) | 48 -> Ok (K Dnskey)
-    | 50 -> Ok (K Nsec3) |  52 -> Ok (K Tlsa) | 257 -> Ok (K Caa)
+    | 50 -> Ok (K Nsec3) | 29 -> Ok (K Loc) | 52 -> Ok (K Tlsa)
```
hannesm, line 2382: same, please keep the order of integers.

RyanGibb:
Thanks for the comments @hannesm! Apologies for the erroneous files. The `.envrc` and `shell.nix` where for making opam work with nix nicely in my local environment as outlined [here](https://github.com/ocaml/opam/issues/5124#issuecomment-1119740117).

src/dns.mli
```diff
@@ -791,6 +791,54 @@ module Edns : sig
...
+    -> longitude:((int32 * int32 * int32) * bool)
+    -> altitude:int64
+    -> precision:(int64 * int64 * int64)
+    -> t
+  (** [parse ~latitude ~longitude ~altitude ~precision] Parse a human-readable format
+     to a Loc record.
+
+     [latitude] is represented by
+       ((degress, mintues, seconds), direction),
+     where direction is true for North and false for South.
```
hannesm, line 821: Thanks for your documentation. I'd find it slightly easier to use for direction a polymorphic variant, `` [ `North | `South ] `` (resp. `` [ `West | `East ] `` below). Or is that inconvenient in using this function?

RyanGibb: That sounds like a great idea. I will investigate implementing this.

RyanGibb: @hannesm, I've added this now!

src/dns.ml
```diff
@@ -1303,6 +1303,180 @@ module Nsec3 = struct
...
+        let ( / ) = Int64.div in
+        let m = p / (pow10 exponent) in
+        if m > 9L then 9 else Int64.to_int m
+      in
+      (Int.shift_left mantissa 4) lor exponent
+    in
+    (encode size, encode horiz_pre, encode vert_pre)
+  
+  (* RFC 1876 Appendix A *)
+  let parse ~latitude:latitude ~longitude:longitude ~altitude:altitude ~precision:precision =
```
hannesm, line 1354:
```suggestion
  let parse ~latitude ~longitude ~altitude ~precision =
```

src/dns.ml
```diff
@@ -1303,6 +1303,180 @@ module Nsec3 = struct
...
+
+  let compare a b =
+    List.fold_right andThen [
+      compare a.latitude b.latitude ;
+      compare a.longitude b.longitude ;
+      compare a.altitude b.altitude ;
+      compare a.size b.size ;
+      compare a.horiz_pre b.horiz_pre ;
+      compare a.vert_pre b.vert_pre ;
+    ] 0
```
hannesm, line 1457: these use the polymorphic compare function all over... and could just be written as `compare a b` -- but the polymorphic compare isn't nice... so can we use `int32_compare` (defined in `dns.ml`) and `int_compare` respectively?

src/dns.ml
```diff
@@ -1303,6 +1303,180 @@ module Nsec3 = struct
...
+        [alt_print loc.altitude] @ (List.map precision_print [loc.size; loc.horiz_pre; loc.vert_pre])    
+      )
+    in
+    String.concat " " ([lat_string; long_string;] @ meter_values)
+
+  let pp ppf loc = Fmt.pf ppf "LOC %s" (to_string loc)
+
+  let compare a b =
+    List.fold_right andThen [
+      compare a.latitude b.latitude ;
```
hannesm, line 1451:
```suggestion
      int32_compare a.latitude b.latitude ;
```

src/dns.ml
```diff
@@ -1303,6 +1303,180 @@ module Nsec3 = struct
...
+      compare a.longitude b.longitude ;
```
hannesm, line 1452:
```suggestion
      int32_compare a.longitude b.longitude ;
```

src/dns.ml
```diff
@@ -1303,6 +1303,180 @@ module Nsec3 = struct
...
+      compare a.altitude b.altitude ;
```
hannesm, line 1453:
```suggestion
      int32_compare a.altitude b.altitude ;
```

src/dns.ml
```diff
@@ -1303,6 +1303,180 @@ module Nsec3 = struct
...
+      compare a.size b.size ;
+      compare a.horiz_pre b.horiz_pre ;
```
hannesm, line 1455:
```suggestion
      int_compare a.horiz_pre b.horiz_pre ;
```

src/dns.ml
```diff
@@ -1303,6 +1303,180 @@ module Nsec3 = struct
...
+      compare a.vert_pre b.vert_pre ;
```
hannesm, line 1456:
```suggestion
      int_compare a.vert_pre b.vert_pre ;
```

src/dns.ml
```diff
@@ -2212,9 +2390,10 @@ module Rr_map = struct
   let of_int ?(off = 0) = function
     | 1 -> Ok (K A) | 2 -> Ok (K Ns) | 5 -> Ok (K Cname) | 6 -> Ok (K Soa)
     | 12 -> Ok (K Ptr) | 15 -> Ok (K Mx) | 16 -> Ok (K Txt) | 28 -> Ok (K Aaaa)
-    | 33 -> Ok (K Srv) | 43 -> Ok (K Ds) | 44 -> Ok (K Sshfp)
+    | 29 -> Ok (K Loc) | 33 -> Ok (K Srv) | 43 -> Ok (K Ds) | 44 -> Ok (K Sshfp)
     | 46 -> Ok (K Rrsig) | 47 -> Ok (K Nsec) | 48 -> Ok (K Dnskey)
-    | 50 -> Ok (K Nsec3) |  52 -> Ok (K Tlsa) | 257 -> Ok (K Caa)
+    | 50 -> Ok (K Nsec3) | 52 -> Ok (K Tlsa)
+    | 257 -> Ok (K Caa)
```
hannesm, line 2396:
```suggestion
```

src/dns.ml
```diff
@@ -3542,7 +3751,7 @@ module Packet = struct
       *)
       let* () =
         guard (ancount >= 1)
-          (`Malformed (6, Fmt.str "AXFR needs at least one RRs in answer %d" ancount))
+          (`Malformed (6, Fmt.str "AXFR needs at l`east one RRs in answer %d" ancount))
```
hannesm, line 3754:
```suggestion
          (`Malformed (6, Fmt.str "AXFR needs at least one RRs in answer %d" ancount))
```

src/dns.ml
```diff
@@ -3702,7 +3911,7 @@ module Packet = struct
       let* () = guard (not (Flags.mem `Truncation flags)) `Partial in
       let* () =
         guard (ancount >= 1)
-          (`Malformed (6, Fmt.str "IXFR needs at least one RRs in answer %d" ancount))
+          (`Malformed (6, Fmt.str "IXFR needs at l`east one RRs in answer %d" ancount))
```
hannesm, line 3914:
```suggestion
          (`Malformed (6, Fmt.str "IXFR needs at least one RRs in answer %d" ancount))
```

src/dns.mli
```diff
@@ -791,6 +791,49 @@ module Edns : sig
...
+    horiz_pre : int;
+    vert_pre : int;
+  }
+  (** The type of a Loc record. *)
+
+  val parse :
+    latitude:((int32 * int32 * int32) * [ `North | `South ])
+    -> longitude:((int32 * int32 * int32) * [ `East | `West ])
+    -> altitude:int64
+    -> precision:(int64 * int64 * int64)
```
hannesm, line 814: it is likely perfectly fine, but why is precision here a triple of int64, above in the type a triple of int? is this fine for 32 bit systems?

hannesm:
thank you for your contribution. I added minor comments and suggestions above. IMHO this is mostly ready for being merged.

src/dns.ml
RyanGibb: whoops! Thanks for catching this

src/dns.ml
hannesm: oh thanks, I opened #314 to avoid polymorphic compare usage.

src/dns.mli
RyanGibb: The precision values (size, horizontal precision, and vertical precision) are they are represented on-the-wire with a 4 bit mantissa and 4 bit exponent, limited to 9^9 according to the RFC. However, this encoding method can physically represent values up to 15^15, larger than what can fit in an int32. A 64 bit representation allows us to handle this case, despite the fact that it is undefined according to the RFC.

We use an int64 for parsing the altitude as it's an unsigned 32 bit number.  A number from 2^31 to 2^32 failes to be parsed by `Int32.of_string`. This is converted to an `int32` for as on the wire representation in `alt_parse`. Note that it requires conversion to an unsigned number when decoding in `alt_print`.
 
 The latitude and longitude store the degrees, minutes, and seconds of an arc. All fit well within an int32.
 
 So, all is fine for 32-bit systems.

RyanGibb:
Thank you for the suggestions, and for all the help getting this here!

src/dns.ml
```diff
@@ -1303,6 +1303,180 @@ module Nsec3 = struct
...
+      )
+    in
+    String.concat " " ([lat_string; long_string;] @ meter_values)
+
+  let pp ppf loc = Fmt.pf ppf "LOC %s" (to_string loc)
+
+  let compare a b =
+    List.fold_right andThen [
+      int32_compare a.latitude b.latitude ;
+      int32_compare a.longitude b.longitude ;
```
hannesm, line 1452:
```suggestion
      Int32.compare a.longitude b.longitude ;
```

src/dns.ml
```diff
@@ -1303,6 +1303,180 @@ module Nsec3 = struct
...
+      int32_compare a.altitude b.altitude ;
```
hannesm, line 1453:
```suggestion
      Int32.compare a.altitude b.altitude ;
```

src/dns.ml
```diff
@@ -1303,6 +1303,180 @@ module Nsec3 = struct
...
+      int_compare a.size b.size ;
```
hannesm, line 1454:
```suggestion
      Int.compare a.size b.size ;
```

src/dns.ml
```diff
@@ -1303,6 +1303,180 @@ module Nsec3 = struct
...
+      int_compare a.horiz_pre b.horiz_pre ;
```
hannesm, line 1455:
```suggestion
      Int.compare a.horiz_pre b.horiz_pre ;
```

hannesm:
sorry, in #314 I decided it doesn't make much sense to define `int_compare` / `int32_compare` in dns.ml, and got rid of it.

src/dns.mli
RyanGibb: Oh - sorry - I just realized I may have misinterpreted your question. The reason why the precision values are stored as an `int` in `Loc.t` is because we only need 8 bits - 4 for the mantissa and 4 for the exponent. So all is find so long as the system is 8-bit or more :D

hannesm:
CI is green, I squash-merged.

PR #311, opened 2022-05-04
app/ozone: be more explicit when showing errors

undu:
Now the domain name is stated to be domain name, so it can be related to
the filename; and only records that are not part of the domain are
printed


app/ozone.ml
```diff
@@ -11,14 +12,13 @@ let load_zone zone =
...
-    Domain_name.Map.for_all
-      (fun name _ -> Domain_name.is_subdomain ~domain ~subdomain:name)
+  let bad = Domain_name.Map.filter
+      (fun name _ -> not @@ Domain_name.is_subdomain ~domain ~subdomain:name)
       rrs
   in
-  if not good then
-    Error (`Msg (Fmt.str "an entry of %a is not in its zone, won't handle this@.%a"
-                   Domain_name.pp domain Dns.Name_rr_map.pp rrs))
+  if bad <> Domain_name.Map.empty then
```
reynir, line 19: Rather than using polymorphic compare use `Domain_name.Map.equal` or even better `is_empty`:
```suggestion
  if not (Domain_name.Map.is_empty bad) then
```

reynir (changes requested):
Thanks! Besides a minor nitpick I think this looks good.

app/ozone.ml
```diff
@@ -11,14 +12,13 @@ let load_zone zone =
   let* data = Bos.OS.File.read Fpath.(v zone) in
   let* rrs = Dns_zone.parse data in
   let domain = Domain_name.of_string_exn Fpath.(basename (v zone)) in
-  let good =
-    Domain_name.Map.for_all
-      (fun name _ -> Domain_name.is_subdomain ~domain ~subdomain:name)
+  let bad = Domain_name.Map.filter
+      (fun name _ -> not @@ Domain_name.is_subdomain ~domain ~subdomain:name)
```
hannesm, line 16:
```suggestion
      (fun name _ -> not (Domain_name.is_subdomain ~domain ~subdomain:name))
```

hannesm (approved):

reynir (approved):

PR #312, opened 2022-06-23
Add eio as one of the backends for ocaml-dns client

bikallem:
This PR adds support for eio backend for ocaml-dns client. At the moment it only adds `tcp` tansport protocol support. I intend to add `tls` or DNS over TLS support in the future once eio support for ocaml-tls is done.

- [ ] `Transport` module
     - use Eio.Mutex around `Queue` calls ? (Not sure if it is needed yet)
- [ ] Integrate `mirage-crypto-rng-eio` (https://github.com/mirage/mirage-crypto/pull/155)
- [ ] `ohost` executable execising `dns-client-eio` package (https://github.com/bikallem/ocaml-dns/blob/main/unix/client/ohost.ml)

This PR is in a draft state as it depends on a few unreleased packages:
1. A new version of eio with the Mutex support
2. `mirage-crypto-rng-eio` 
3. At the moment `eio` doesn't support monotonic clock, so we still use `mtime`. It may be preferable to use monotonic clock support from `eio` once support for it lands in eio. https://github.com/ocaml-multicore/eio/issues/229


talex5 (commented):
I experimented with some API changes here: https://github.com/bikallem/ocaml-dns/compare/eio...talex5:ocaml-dns:eio?expand=1

The main ones are:

1. Instead of providing a `run` function that just calls mirage-crypto, it's just as easy for the user call it themselves (and better if they're using crypto for other purposes).
2. There's no need to pass `Client` as a first-class module. We know what it is statically.
3. I moved `create` to `create_from_env` for people who want that, and added a more explicit `create` function so people can pass the inputs separately.

This changes the user code from:
```ocaml
  Eio_main.run @@ fun env -> 
  Eio.Switch.run @@ fun sw -> 
  Dns_client_eio.run env @@ fun (module Client) ->   
  let env = (env :> Dns_client_eio.env) in
  let c = Client.create (env, sw) in
```
to
```ocaml
  Eio_main.run @@ fun env -> 
  Mirage_crypto_rng_eio.run (module Mirage_crypto_rng.Fortuna) env @@ fun () ->
  Eio.Switch.run @@ fun sw -> 
  let c = Client.create_from_env ~sw env in
```

The needs to specify `Fortuna` explicitly seems a bit ugly to me, but if we don't want that then it should be fixed in mirage-crypto.

I'm not sure about passing `resolv.conf` as an `Eio.Path`. Possibly it should just be a function `unit -> string`. Also, Eio should provide some convenience types for paths (maybe `Path.rw` and `Path.ro` or something).

bikallem:
I have rebased/updated the PR to address reviewer feedback.

>(Was the idea of the first class module to force users to initialise mirage-crypto before using the library? If so, we could probably fix this more easily by having Mirage_crypto_rng_eio.run return a token and passing that to create_from_env. Passing first-class modules around tends to get awkward.)

Indeed, the idea of the API was to maintain the invariant that that consumers of the library to be able to use the module only when run under `Mirage_crypto_rng_eio.run`. The current API aims to maintain that invariant as well as be convenient such that one doesn't have to manually run (although they can if they so desire) the mirage crypto rng generator.

bikallem:
Hmmm ... not sure why the ci is failing. Can anyone help?

talex5:
It says:

> lru.0.3.0: Requires ocaml >= 4.03.0 & < 5.0

hannesm:
I'm not particularly a fan of passing first-class modules around, and don't think there's a good reason here.

I wonder what the "EIO" and "RANDOM" story is (if there's any):
- I have seen some work on mirage-crypto-rng-eio, so you can initialize and feed the Fortuna RNG;
- There has been as well some support for getrandom (but seeing the discussion of https://github.com/ocaml-multicore/eio/pull/344 I'm not sure whether this can be safely used) -- this is exposed as "secure_random" by the "StdEnv" AFAICT (though I'm not sure what else is in this "StdEnv" and why);
- And OCaml in 5.0 has support for getentropy to feed its internal RNG (Random module).

So, what is the plan for EIO and RANDOM? How and who should decide where to take random numbers from? Is this all up to the user (and do they have a unified interface)? In DNS, we don't need cryptographically secure random, something that is not predictable is sufficient (it's used mainly for the ID anyways, to avoid spoofing of DNS packets (see https://en.wikipedia.org/wiki/DNS_spoofing if interested in details)).

talex5:
> I'm not particularly a fan of passing first-class modules around, and don't think there's a good reason here.

Indeed. I removed it in my branch (see https://github.com/mirage/ocaml-dns/pull/312#pullrequestreview-1088918946).

> (but seeing the discussion of https://github.com/ocaml-multicore/eio/pull/344 I'm not sure whether this can be safely used)

The PR is just trying to make it harder to misuse the API. mirage-crypto-rng-eio already uses it correctly (I'm not aware of anyone using it incorrectly).

> In DNS, we don't need cryptographically secure random, something that is not predictable is sufficient

I assume it's just copying the existing lwt code, which uses `mirage-crypto-rng.lwt`. Would be good to reduce the dependencies if it's not needed :-)

hannesm:
Trying once more. In eio, do you have a user guide / approach to randomness?

I can see three different things going on here:
- the OCaml 5 runtime (Stdlib.Random -- which is a LXM: Better Splittable Pseudorandom Number
Generators (and Almost as Fast); seeding is done via `getentropy()` if available)
- the eio providing a "secure_random" in "StdEnv", which calls out to `getrandom()`
- mirage-crypto-rng-eio calling out to rdrand/rdseed and getrandom/getentropy (periodic task) feeding a Fortuna RNG

Now, my questions are:
- is there a unified interface between these implementations?
- is there an eio-recommended usage of either of these?
- is the approach to have one syscall (getentropy/getrandom) each time some random is needed too expensive?

Maybe this PR here is the wrong context for these questions and considerations, but for me it is important to understand how "eio" should be used in order to avoid having to back out such changes.

bikallem:
I have now removed the use of first class modules. the `ohost.exe` included in the PR don't seem to be working after latest rebase and commit, so I debugging through the issue.

bikallem:
> Trying once more. In eio, do you have a user guide / approach to randomness?
> 
> I can see three different things going on here:
> 
>     * the OCaml 5 runtime (Stdlib.Random -- which is a LXM: Better Splittable Pseudorandom Number
>       Generators (and Almost as Fast); seeding is done via `getentropy()` if available)
> 
>     * the eio providing a "secure_random" in "StdEnv", which calls out to `getrandom()`
> 
>     * mirage-crypto-rng-eio calling out to rdrand/rdseed and getrandom/getentropy (periodic task) feeding a Fortuna RNG
> 
> 
> Now, my questions are:
> 
>     * is there a unified interface between these implementations?
> 
>     * is there an eio-recommended usage of either of these?
> 
>     * is the approach to have one syscall (getentropy/getrandom) each time some random is needed too expensive?
> 
> 
> Maybe this PR here is the wrong context for these questions and considerations, but for me it is important to understand how "eio" should be used in order to avoid having to back out such changes.

All valid and good questions and for which I do not have quick answers. Perhaps they may get a better traction as an issue in the eio repo?

eio/client/dns_client_eio.ml
```diff
@@ -0,0 +1,204 @@
...
+
+  let read_file file =
+    match Eio.Path.load file with
+    | content -> Ok content
+    | exception e ->
+      Fmt.error_msg "Error while reading file %a: %a" Eio.Path.pp file Fmt.exn e
+
+  (* Prioritises IPv6 nameservers before IPv4 nameservers so that we 
+     are more conformant with the happy eyballs RFC when implementing it. 
+     https://datatracker.ietf.org/doc/html/rfc8305#section-3  *)
```
hannesm, line 49: I'm curious, why not just use the happy-eyeballs library, as done in dns-client for lwt and mirage?

eio/client/dns_client_eio.ml
```diff
@@ -0,0 +1,204 @@
...
+  type nonrec stack = stack
+  type +'a io = 'a
+  type context = <Eio.Net.stream_socket; Eio.Flow.close>
+
+  type nameservers = 
+    | Static of io_addr Queue.t
+    | Resolv_conf of {
+        mutable nameservers : io_addr Queue.t ;
+        mutable digest      : Digest.t option ;
+      }
```
hannesm, line 32: what is the purpose of the Queue.t sprinkled here?

eio/client/dns_client_eio.ml
bikallem: For now all I wanted to do was sort name-server addresses in IP6 order. However, this package `dns-client-eio` will be required to implement `happy-eyeballs-eio`. So I want to leave happy-eyeballs implementation to the latter package.

eio/client/dns_client_eio.ml
```diff
@@ -0,0 +1,204 @@
+type 'a env = <
+  clock         : Eio.Time.clock ;
+  net           : Eio.Net.t ;
+  fs            : Eio.Fs.dir Eio.Path.t ;
+  secure_random : Eio.Flow.source;
+  ..
+> as 'a
+
+type io_addr = Ipaddr.t * int
```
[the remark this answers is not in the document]
bikallem: At the moment it only support `Tcp. I am looking at the `UDP support separately after resolving this issue in `eio` (https://github.com/ocaml-multicore/eio/issues/345). `tls` not too sure at this stage.

eio/client/dns_client_eio.ml
bikallem: `Queue.t` is used to implement round-robin attempt at connecting to given nameservers. Basically if an attempt to connect to a namesever fails then it is pushed to the end of the queue and we try the next one in the queue. Ordered as IPv6 ascending. https://github.com/mirage/ocaml-dns/pull/312/files#diff-f1c4d1e041cdee765597667b7e7a51894b20b99d1f5deb97adf94925f82d9c1aR152

eio/client/dns_client_eio.ml
hannesm: The happy-eyeballs package is independent of dns -> the happy-eyeballs-lwt/mirage/eio are dependent on dns-client. The pure happy-eyeballs is especially crafted for the use in dns, it allows to deal with ipv4 and ipv6 addresses. Please take a look into dns-client-lwt/dns-client-mirage, there's no reason to ad-hoc "sort name-server addresses in IP6 order".

talex5:
> Trying once more. In eio, do you have a user guide / approach to randomness?

No. At the moment, Eio is just exposing what the OS provides: e.g. Linux provides `getrandom(2)` and Eio provides an OCaml API for using that.

[ mirage-crypto, Eio's secure_random, stdlib random ]

> is there a unified interface between these implementations?

Not at the moment. Eio's `secure_random` is just an `Eio.Flow.source` (byte stream). It would be possible to access the other two with the same interface if desired, though.

> is there an eio-recommended usage of either of these?

I was imagining that:
- If you want secure random numbers, you use mirage-crypto (which uses Eio to call `getrandom` as needed).
- If you don't, you use `Stdlib.Random` (possibly seeded using Eio, but probably using OCaml's built-in seeding).

> is the approach to have one syscall (getentropy/getrandom) each time some random is needed too expensive?

I was expecting `secure_random` it to be used only for seeding, but depending on the requirements I guess it could be fine to use it directly.

The API docs should probably mention that it just wraps the OS syscall and may not be particularly fast. Or perhaps users could wrap it with `Eio.Buf_read` to read larger amounts of random data in chunks.

You know a lot more about RNGs than me - how would you like it to work?

eio/client/dns_client_eio.ml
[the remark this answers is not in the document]
bikallem: Waiting for this to merge https://github.com/ocaml-multicore/eio/pull/360

haesbaert:
I feel two main points that Hannes is concerned are:
a) Randomness
b) The queue that tries to work-around the lack of happy-meatballs

I think **a** is a non-issue, as @hannesm said, we just want random 16bit ids that do not need to be cryptographically secure.
In EIO, as @talex5 mentioned, we provide a `secure_random` where you request cryptographically secure entropy, so this port could use `secure_random` and use it to retrieve "more often than not" unique 16bit ids. The other alternative is to call the stdlib Random, which for this case would be enough.
 I agree with @talex5, if you deem that retrieving that entropy is too expensive, that's on you to pull larger chunks and deplete them when you need. EIO is not keeping state, we're just tapping into whatever the OS thinks is "good entropy". Personally I'd use it directly until I see 1 brazilion packets per second being hindered by it. Worth mentioning that getentropy/getrandom guarantee to not block on <= 256 bytes requests (it's actually 4k for getrandom, the manpage is lying).


Maybe we should attack **b** first, so that this port doesn't look too especial. In regards to functionality, the EIO port should "look a lot" like the Lwt in terms of structure and primitives, and perhaps if we work on happy-balls first this becomes a bit more of a 1:1 port. I had a quick look at the Lwt so forgive me for any misunderstandings.

eio/client/dns_client_eio.ml
bikallem: Ack. I will look into using happy-eyeballs for this part.

bikallem:
Needs https://github.com/mirleft/ocaml-tls/pull/458 to operate correctly.

hannesm:
So it seems we're a bit stuck on this PR:
- (a) there's some lack of a "random" story for eio (and it seems nobody is pushing that forward (no, it won't be me))
- (b) concurrent reads and writes on TLS (further discussed in https://github.com/mirleft/ocaml-tls/issues/464, still without a solution)

How to move forward here? I unlikely have time before Q3 2023 to look into "eio", but I suspect you'd like to have "something merged" sooner. But I'm still overwhelmed by the complexity of eio and don't feel confident to grasp the semantics of what is happening where (and why concurrently etc.).

bikallem:
@hannesm This PR is now ready. It doesn't have the TLS issue anymore. 

 > (a) there's some lack of a "random" story for eio (and it seems nobody is pushing that forward (no, it won't be me))

The issue is separate from dns-client-eio and should be handled in the eio repo.

dinosaure:
I take a long look about this PR and it seems that DNS request by another domain seems not an option with the current proposed design in this PR - at least, the use of "mutable" without protection makes me think there could be problems. 

The reason I point to this specific usage is that it seems to me that the design of happy-eyeballs (the core package, not the `happy-eyeballs-lwt`) is to be a background task that maintains a "pool" of TCP/IP connections to nameservers and can resolve a DNS request from the user. It is therefore "legitimate" to think that one domain would take care of such a task in the background while the user application could interact with it from another domain to resolve domain-name.

bikallem:
> I take a long look about this PR and it seems that DNS request by another domain seems not an option with the current proposed design in this PR - at least, the use of "mutable" without protection makes me think there could be problems.

In current eio, IIUIC resource/connection/switch created in one domain can't be moved/used in another domain. Therefore, even if happy eyeballs is able to create connections, I am not sure if those connections can be used/shared in an ad-hoc manner by the domains. However, this does not preclude having multiple/parallel dns client requests via OCaml domains - which the current design offers/enables.

PR #315, opened 2022-08-16
make polymorphic variant errors open

copy:
For composing errors. I tried to be thorough, but personally I only care about dns.mli. I skipped `Dns.Tsig_op.e`  because it propagates all the way to `Dns_server.t` (which contains a `Tsig_op.verify`, which returns this error).

### DIFF
--- a/cache/dns_cache.ml
+++ b/cache/dns_cache.ml
@@ -242,10 +242,10 @@ let get_any cache ts name =
 
 let get_or_cname : type a . t -> int64 -> [`raw] Domain_name.t -> a Rr_map.key ->
   t * ([ a entry | `Alias of int32 * [`raw] Domain_name.t] * rank,
-       [ `Cache_drop | `Cache_miss ]) result =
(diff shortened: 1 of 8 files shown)


hannesm:
I'm not convinced this is the right way to go. You can on the calling site do this opening the error message (see e.g. https://github.com/dbuenzli/rresult/blob/v0.7.0/src/rresult.mli#L133-L135 as example, the implementation is the identity). Take the signatures in dns_cache.mli as example: your proposed patch makes them less strict.

copy:
`open_error_msg` only works for ```(_, [`Msg of string]) result```, so ocaml-dns would have to provide its own version for each error type. Is there any way to write this less verbosely?

```ocaml
type myerror = [Dns.Packet.err | `Foo | `Bar]

let (let*) = Result.bind

let main packet : (unit, [> myerror]) result =
  let* _ = (Dns.Packet.decode packet : (_, Dns.Packet.err) result :> (_, [> myerror]) result) in
  let* _ = Error `Foo in
  Ok ()
```

PR #322, opened 2022-10-06
dns-client-mirage udp

hannesm:


mirage/client/dns_client_mirage.ml
```diff
@@ -134,9 +151,19 @@ module Make (R : Mirage_random.S) (T : Mirage_time.S) (M : Mirage_clock.MCLOCK)
       Lwt_condition.wait t.timer_condition >>= fun () ->
       loop ()
 
+    let udp_port = 12345
```
hannesm, line 154: TODO: should this be randomized? We'll need UDP listeners for each port (and ensure that either each is used only once or that unlisten is not done always (if there are two packets from port = x, only after the last packet sent the unlisten may be done)

avsm: [RFC5452](https://www.rfc-editor.org/rfc/rfc5452) has a bunch of advice on this. It's a little scary to use a fixed source port for making DNS requests without DNSSEC verification. Spoofing responses is just so much easier if you can spam them to the same port.

hannesm: yes. refering to all of dan kaminsky work. that's why I put that comment here to remind ourselves to adjust the code before pressing merge and release ;)

hannesm: @haesbaert @avsm please check https://github.com/mirage/ocaml-dns/pull/322/commits/aa7458ad7a697f5a569e17b5abb5e4fdbd28a110 that should fix your reservations.

hannesm:
@reynir if you could take a look at this, that'd be great :)

reynir (commented):
I am a little concerned that `generate_udp_port` doesn't check that nothing is listening on the generated port. The interface of `UDP.listen` always replaces an existing listener. An application that uses DNS and listens on UDP may eventually break because a dns lookup triggered `generate_udp_port` to generate the application port. However, with the current interface I can't see how we can do better.

mirage/client/dns_client_mirage.mli
```diff
@@ -1,25 +1,32 @@
...
-        we will communicate with it {i via} the TCP/IP protocol plus the TLS
+        we will communicate with it {i via} the TCP protocol plus the TLS
         encrypted layer. The user can verify the nameserver {i via} an
         {i authenticator} (see {!X509.Authenticator.of_string} for the format
         of it). The {i hostname} can be provided to be used as peer name by the
         authenticator. By default, {!Ca_certs_nss.authenticator} is used.
     *)
 
   val connect :
+    ?size:int ->
```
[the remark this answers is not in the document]
hannesm: the issue is that `?size` is used in dns_client.mli -- so maybe we should rename all the `?size` to `?cache_size`. (I'll do that).

mirage/client/dns_client_mirage.ml
```diff
@@ -289,7 +390,7 @@ module Make (R : Mirage_random.S) (T : Mirage_time.S) (M : Mirage_clock.MCLOCK)
                 connect_ns t t.nameservers >|= function
                 | Error `Msg msg ->
                   Log.err (fun m -> m "error while connecting to resolver: %s" msg)
-                | Ok () -> ()
+                | Ok _ -> ()
```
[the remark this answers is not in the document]
hannesm: agreed, that was a temporary change for when it did not return `()`, which I failed to revert. thanks! :)

mirage/client/dns_client_mirage.ml
```diff
@@ -320,25 +421,49 @@ module Make (R : Mirage_random.S) (T : Mirage_time.S) (M : Mirage_clock.MCLOCK)
...
+      | `Udp -> Lwt.return (Ok (`Udp, t))
+      | `Tcp -> match t.flow, t.connected_condition with
+        | Some _, _ -> Lwt.return (Ok (`Tcp, t))
+        | None, Some w ->
+          Lwt_condition.wait w >>= fun () ->
+          connect t
+        | None, None ->
+          connect_ns t t.nameservers >|= function
+          | Ok () -> Ok (`Tcp, t)
+          | Error `Msg msg -> Error (`Msg msg)
```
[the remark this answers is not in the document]
hannesm: I guess #286, and indeed in mirleft it looked like the unbounded queue of outstanding requests were pretty hard on the poor tp link 4g modem once one tcp session failed, and the stub-resolve re-requested everything outstanding. maybe in such a case, the requests should just be dropped (given the behaviour of DNS clients to re-request if they're really interested anyways).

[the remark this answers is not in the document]
hannesm: also opened #323 to track the issue mentioned above

mirage/client/dns_client_mirage.mli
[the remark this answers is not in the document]
hannesm: see 6333f96

mirage/client/dns_client_mirage.ml
```diff
@@ -320,25 +421,49 @@ module Make (R : Mirage_random.S) (T : Mirage_time.S) (M : Mirage_clock.MCLOCK)
...
+          in
+          let id = Cstruct.BE.get_uint16 tx 0 in
+          with_timeout t.timeout_ns
+            (let open Lwt_result.Infix in
+             (let open Lwt.Infix in
+              let udp_port = generate_udp_port t in
+              S.UDP.listen (S.udp t.stack) ~port:udp_port (read_udp t udp_port dst);
+              S.UDP.write ~src_port:udp_port ~dst ~dst_port (S.udp t.stack) tx
+              >|= function
+              | Error e -> Error (`Msg (Fmt.to_to_string S.UDP.pp_error e))
```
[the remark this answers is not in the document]
hannesm: done in 38878b6430aa602d6d30057c82c9c8dac784511a

mirage/client/dns_client_mirage.ml
```diff
@@ -134,9 +196,37 @@ module Make (R : Mirage_random.S) (T : Mirage_time.S) (M : Mirage_clock.MCLOCK)
       Lwt_condition.wait t.timer_condition >>= fun () ->
       loop ()
 
+    let read_udp t port ip ~src ~dst:_ ~src_port data =
+      (* TODO: compare dst being us *)
+      if (port = src_port && Ipaddr.compare ip src = 0) ||
+         (src_port = t.last_udp_port && Ipaddr.(compare ip (V4 V4.any) = 0))
+      then
+        (let id = Cstruct.BE.get_uint16 data 0 in
```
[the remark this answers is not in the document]
hannesm: done in 3a215dcf23e04a0ec12b216fd3d588ee6ca0a5fb

hannesm:
thanks for your review -- I addressed your comments :)

hannesm:
the interface of UDP.S with listen and unlisten is not optimal, indeed -- and should be enhanced -- maybe as simple as providing a `free_port` in the interface.

mirage/client/dns_client_mirage.ml
```diff
@@ -192,10 +192,10 @@ The format of a nameserver is:
       Lwt_condition.wait t.timer_condition >>= fun () ->
       loop ()
 
-    let read_udp t port ip ~src ~dst:_ ~src_port data =
+    let read_udp t port ip ~src ~dst:_ ~src_port:_ data =
       (* TODO: compare dst being us *)
-      if (port = src_port && Ipaddr.compare ip src = 0) ||
-         (src_port = t.last_udp_port && Ipaddr.(compare ip (V4 V4.any) = 0)) &&
+      if (Ipaddr.compare ip src = 0 ||
+          (port = t.last_udp_port && Ipaddr.(compare ip (V4 V4.any) = 0))) &&
```
reynir, line 198: In all call sites, we pass `dst_port` to `read_udp` as `port` when setting up the listener. So the `port = dst_port` should be unnecessary.
I'm not sure about comparing `port` with `t.last_udp_port`. In `create` we update `t` with `last_udp_port` to the generated port. In the other callsite (`send_recv`) we *don't* update `last_udp_port`. With that in mind I'm confused what it is we want with `last_udp_port` which to me seems to be in fact the first udp port. It also seems to me that we will never receive answers unless we happen to send and receive on the port generated in `create`.

reynir: I am now even more confused with the `V4.any` bit. If `generate_udp_port` fails to generate a port and reuses `last_udp_port` we will be overriding the `V4.any` listener in `send_recv`. I'm not sure what the purpose is in setting up the listener in `create` then.

hannesm: Sorry for the confusion. The reasoning is as follows:
- the resource of UDP porrts is limited to 2^16 - 1024 (since DNS servers dislike being asked from a privileged (<1024) port)
- we would like to have as much entropy as possible to prevent DNS spoofing (https://en.wikipedia.org/wiki/DNS_spoofing), namely: a random DNS identifier, random UDP source port
- we may end up in a situation that all UDP ports have been used (when we have 2^16 - 1024 requests concurrently)

In general, we allocate a not-yet used UDP port in `send_recv`, listen (with `read_udp`) on it, and unlisten once a reply or timeout happened.

Maybe the `last_udp_port` is over-engineered, but the idea is: if there's no free UDP port, use a port reserved at startup (in `create`) to be able to send requests and receive replies. Since I wrote this, I think this is overly complex and when we're out of UDP ports we're in trouble anyways. I'll push a subsequent commit that removes this complexity.

reynir (approved):

mirage/client/dns_client_mirage.ml
palainp: Hi, do you think this reasoning  is also relevant for qubes-mirage-firewall?

hannesm: In qubes-mirage-firewall we only ever talk to the qubesdb-provided dns resolver, and use by default a random UDP port (but still have a `last_resort_port` allocated at start). I don't think we need to change anything in qubes-mirage-firewall (though we could remove the `last_resort_port` and fail if there's no free UDP port -- the downside is that DNS resolution may be required for rule matching, and the qubes-mirage-firewall does NAT also for UDP, thus more UDP ports are used).

PR #327, opened 2022-11-24
Upgrade to dune 2.0

reynir:
I found that dune would attempt to compile ocaml-dns to javascript if its source was included as a subdirectory in happy-eyeballs. With help from people on the discord I had two solutions, and one of them was to upgrade to dune-lang 2.0.


hannesm:
ok, but why then `2.0.1`? to me, `2.0` or `2.0.0` would be more intuitive (dune 2.0.0 is also marked as unavailable).

reynir:
Sure. I used 2.0.1 as 2.0.0 was unavailable. I'll change it.

dune-project
```diff
@@ -1,2 +1,2 @@
-(lang dune 1.2)
+(lang dune 2.0)
 (name dns)
```
hannesm, line 2:
```suggestion
(name dns)
(formatting disabled)
```

Dune 2.0 imposes some ocamlformat stuff, which we're not using here (yet?).

reynir: Ah, I wasn't aware. Maybe we should?

hannesm:
thanks

PR #329, opened 2022-12-01
adapt to new happy eyeballs

hannesm:
goes well with https://github.com/roburio/happy-eyeballs/pull/30


reynir:
I implemented cancellation here as well, but with `Lwt.cancel`. We should apply the same `Lwt.pick` change here as well.

hannesm (commented):
Looks fine, I reviewed https://github.com/mirage/ocaml-dns/pull/329/files?w=1 to avoid whitespace changes.

PR #334, opened 2023-02-15
Resolv.conf parser: allow zone idx (RFC 4007) be present for IPv6 addresses

hannesm:
Inspired by the bug report in #328 by @bikallem, test case taken from there

@bikallem please let me know what you think of this PR, thanks.


bikallem:
LGTM.

PR #336, opened 2023-02-27
A failing test case since #310 was merged

hannesm:
Since I'm a bit short on time, I wonder whether reverting #310 and cutting a release is the path forward. Or, @RyanGibb are you still using LOC records and interesting in them? If yes, would you mind to take a look at the lexer/parser issue and develop a fix (where this test case is passing)?

### DIFF
--- a/test/server.ml
+++ b/test/server.ml
@@ -1625,6 +1625,7 @@ e	A	1.2.3.4
 2m	A	1.2.3.4
 2	A	1.2.3.4
 -2	A	1.2.3.4
+TYPE23	A	1.2.3.4
 |}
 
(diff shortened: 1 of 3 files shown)


hannesm:
Nevermind, found the missing bit in the zone parser (adding LAT_DIR and LONG_DIR to the keyword_or_number rule.

reynir (approved):

hannesm:
I figured there were cases for DS and CAA missing, added them and a more complete test.

hannesm:
and now as well for the remaining tokens `METER` and `NEG_NUMBER`.

PR #337, opened 2023-02-28
Minor correction of the dependency constraints

kit-ty-kate:
Given `dns {= version}` is already there, this doesn't change the behaviour at all but simply makes it clearer


hannesm:
oh thank you, not sure why I introduced a >= there.

PR #338, opened 2023-03-22
add NULL record

RyanGibb:
See https://www.rfc-editor.org/rfc/rfc1035#section-3.3.10

Adds a dependency on `hex` to `dns` for pretty printing hex

### DIFF
--- a/src/dns.ml
+++ b/src/dns.ml
@@ -1480,6 +1480,25 @@ module Loc = struct
 
 end
 
+(* Null record *)
+module Null = struct
+  type t = Cstruct.t
+
+  let pp ppf null = Fmt.pf ppf "NULL %a" Cstruct.hexdump_pp null
+
+  let compare = Cstruct.compare
+
+  let decode names buf ~off ~len =
+    let sub = Cstruct.sub buf off len in
+    Ok (sub, names, off + len)
+
+  let encode null names buf off =
+    let max_len = 65535 in
+    let len = min max_len (Cstruct.length null) in
+    Cstruct.blit null 0 buf off len ;
+    names, off + len
+end
+
 (* certificate authority authorization *)
 module Caa = struct
   type t = {
@@ -2295,6 +2314,7 @@ module Rr_map = struct
   module Ds_set = Set.Make(Ds)
   module Rrsig_set = Set.Make(Rrsig)
   module Loc_set = Set.Make(Loc)
+  module Null_set = Set.Make(Null)
 
   module I : sig
     type t
@@ -2332,6 +2352,7 @@ module Rr_map = struct
     | Nsec : Nsec.t with_ttl rr
     | Nsec3 : Nsec3.t with_ttl rr
     | Loc : Loc_set.t with_ttl rr
+    | Null : Null_set.t with_ttl rr
     | Unknown : I.t -> Txt_set.t with_ttl rr
 
   module K = struct
@@ -2358,6 +2379,7 @@ module Rr_map = struct
       | Nsec, Nsec -> Eq | Nsec, _ -> Lt | _, Nsec -> Gt
       | Nsec3, Nsec3 -> Eq | Nsec3, _ -> Lt | _, Nsec3 -> Gt
       | Loc, Loc -> Eq | Loc, _ -> Lt | _, Loc -> Gt
+      | Null, Null -> Eq | Null, _ -> Lt | _, Null -> Gt
       | Unknown a, Unknown b ->
         let r = I.compare a b in
         if r = 0 then Eq else if r < 0 then Lt else Gt
@@ -2390,6 +2412,7 @@ module Rr_map = struct
     | Nsec, (_, ns), (_, ns') -> Nsec.compare ns ns' = 0
     | Nsec3, (_, ns), (_, ns') -> Nsec3.compare ns ns' = 0
     | Loc, (_, loc), (_, loc') -> Loc_set.equal loc loc'
+    | Null, (_, null), (_, null') -> Null_set.equal null null'
     | Unknown _, (_, data), (_, data') -> Txt_set.equal data data'
 
   let equalb (B (k, v)) (B (k', v')) = match K.compare k k' with
@@ -2397,7 +2420,7 @@ module Rr_map = struct
     | _ -> false
 
   let to_int : type a. a key -> int = function
-    | A -> 1 | Ns -> 2 | Cname -> 5 | Soa -> 6 | Ptr -> 12 | Mx -> 15
+    | A -> 1 | Ns -> 2 | Cname -> 5 | Soa -> 6 | Null -> 10 | Ptr -> 12 | Mx -> 15
     | Txt -> 16 | Aaaa -> 28  | Loc -> 29 | Srv -> 33 | Ds -> 43
     | Sshfp -> 44 | Rrsig -> 46 | Nsec -> 47 | Dnskey -> 48 | Nsec3 -> 50
     | Tlsa -> 52 | Caa -> 257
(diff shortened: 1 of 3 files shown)


hannesm:
Thanks for your contribution. The only question I have is why. Do you actually have a usecase for "NULL"? (I'm more interested in keeping the code base small than to support all resource records we can find.)

RyanGibb:
Yes, my use case is transporting arbitrary bytes over DNS. For an example, see https://github.com/yarrick/iodine.

src/dns.ml
```diff
@@ -1480,6 +1480,27 @@ module Loc = struct
...
+module Null = struct
+  type t = bytes
+
+  let to_string null = Hex.show @@ Hex.of_bytes null
+
+  let pp ppf null = Fmt.pf ppf "NULL %s" (to_string null)
+
+  let compare = Bytes.compare
+
+  let decode_exn names buf ~off ~len =
```
avsm, line 1493: Why's a function called `decode_exn` returning a `result` type? Shouldn't it raise an exception?

RyanGibb: It looks like this naming scheme was introduced here: https://github.com/mirage/ocaml-dns/commit/9ad199d8ac01c235c986b43bdee9d4b82ba5b4da

RyanGibb: > [return a result](https://github.com/mirage/ocaml-dns/blob/1d73cfa96db7452e52b49804d546fad55154aca8/src/dns.ml#L3196-L3253)

I think this could be simplified by replacing the `let*` binding over the Result monad with `let` for those functions that don't return an Error. However perhaps we should leave this to a separate PR and stick with the existing scheme for now.

hannesm: indeed, the naming convention is that decoding functions that may raise are postfixed by `_exn` to illustrate that. And indeed, `Cstruct.BE.get_uint32` can raise if the buffer is smaller than offset + 4.

In this case, both `off` and `len` are passed in (checked) to the function, so how would a realistic example look that leads to an exception? As far as I understand the `Cstruct.to_bytes` documentation, the raise may happen if `off` and/or `len` are provided, and the buffer doesn't meet the size.

hannesm:
CI is failing (`Error (warning 8 [partial-match]): this pattern-matching is not exhaustive`) -- did you try to compile the code locally? I'm not keen to add a dependency for `hex` (hint: look at the source, there is already hex decoding and encoding that may be used in `null` records).

The README states rather clearly that legacy/barely used RR types is not dealt with. There is since some time support for `Unknown : I.t -> Txt_set.t with_ttl rr` in the `Rr_map` -- what is the advantage of supporting `NULL` explicitly? (As far as I can tell from this PR, there's no plan to add support for NULL to the zone parser.) Similarly, I wonder what the use of `LOC` is, so far I haven't seen an application that uses it, or did I miss it?

src/dns.ml
hannesm: E.g. `module Ptr` defines a `decode` function since it won't ever raise an exception.

RyanGibb: > In this case, both `off` and `len` are passed in (checked) to the function, so how would a realistic example look that leads to an exception? As far as I understand the `Cstruct.to_bytes` documentation, the raise may happen if `off` and/or `len` are provided, and the buffer doesn't meet the size.

I suppose if the wrong `off` and `len` where passed? I didn't take into account the callee context when writing this. I'm happy to change it if you prefer though.

hannesm: Ok, then maybe take the `callee context` into account? This function is not exposed, and there's only one call site.

The call site does indeed check `Cstruct.length buf - off >= len` (Rr_map.decode), so as mentioned above the buffer will span at least `off + len` bytes.

RyanGibb:
> CI is failing (Error (warning 8 [partial-match]): this pattern-matching is not exhaustive) -- did you try to compile the code locally?

I did, and am using it live now. I'll investigate.

> I'm not keen to add a dependency for hex (hint: look at the source, there is already hex decoding and encoding that may be used in null records).

I'll take a look.

> There is since some time support for Unknown : I.t -> Txt_set.t with_ttl rr in the Rr_map -- what is the advantage of supporting NULL explicitly?

I don't think I can create a NULL record with this, like I can with:
```
Dns.Rr_map.Null_set.singleton (Bytes.of_string reply)
```

And the advantage of NULL over TXT is that it allows binary data of arbitrary length.

> (As far as I can tell from this PR, there's no plan to add support for NULL to the zone parser.)

I'm serving it dynamically at the moment, and don't have a use for reading it from a zonefile. I could take a look at adding it, but I don't think there's support for encoding binary in the current zonefile parsing.

hannesm:
> > There is since some time support for Unknown : I.t -> Txt_set.t with_ttl rr in the Rr_map -- what is the advantage of supporting NULL explicitly?
> 
> I don't think I can create a NULL record with this, like I can with:
> 
> ```
> Dns.Rr_map.Null_set.singleton (Bytes.of_string reply)
> ```
> 
> And the advantage of NULL over TXT is that it allows binary data of arbitrary length.

"Arbitrary".

Wouldn't something like

```
Dns.Rr_map.(singleton (Unknown (Result.get_ok (I.of_int 10))) (Txt_set.singleton "foobar")) 
```

Ah, for encoding purposes you'd need to prefix the length to the value. Indeed, that's a bit inconvenient.

> > (As far as I can tell from this PR, there's no plan to add support for NULL to the zone parser.)
> 
> I'm serving it dynamically at the moment, and don't have a use for reading it from a zonefile. I could take a look at adding it, but I don't think there's support for encoding binary in the current zonefile parsing.

It is fine to not include it in the zone file parser. But there is support for "encoding binary", as suggested in the RFC, using hex encoding (and TYPENNN as type).

But now I've done enough reviewing on this PR, and won't be back for the rest of the week.

RyanGibb:
> I'm fine with merging this, but there are a few things:
> 
>     * code style: the codebase doesn't use `@@` for a reason (hard to read), please remove its usage

I've removed any instances of `@@` in my PR, although the reason I thought it was appropriate was due to a number of other uses [0].

>     * why is the type `bytes` - I find this rather hard to reason about, and would prefer either Cstruct.t or string (since once you have a Null.t in your hand, the value shouldn't change)

I've changed this to a `Cstruct.t`.

>     * the encode: what would happen if a buffer > 65535 is passed? I understand that other code paths (such as Txt) suffer from similar issues, but I'm not very happy about that

```
    let max_len = 65535 in
    let len = min max_len (Cstruct.length null) in
    Cstruct.blit null 0 buf off len ;
```

Would copy the first 65535 octets from `buf`. We could throw an error instead.

>     * as already mentioned, adding a new dependency on `hex` isn't the path forward, we should drop dependencies, not add them. There's no need for a `to_string` function, and `pp` could use `Cstruct.hexdump_pp`

Done!

>     * the documentation heading is `Text records`, it should be something more related to null!?

And also fixed.

<details>
  <summary>[0] Instances of @@</summary>

```
$ grep -r @@
app/odns.ml:          String.concat " " (List.rev @@ String.sub hex n (hlen-n)::acc)
test/tests.ml:    let encoded = fst @@ encode `Udp res in
test/tests.ml:                (Ok res) (decode @@ encoded))
test/tests.ml:                (Ok res) (decode @@ fst @@ encode `Udp res))
test/tests.ml:                data' (fst @@ encode `Tcp res));
test/tests.ml:    let _ = Format.printf "%a\n" Cstruct.hexdump_pp (fst @@ encode `Udp res) in
test/tests.ml:    let _ = Format.printf "%a" Cstruct.hexdump_pp (fst @@ encode `Udp res) in
test/tests.ml:    Alcotest.(check (result t_ok p_err) "Loc encodes" (Ok res) (decode @@ fst @@ encode `Udp res))
test/client.ml:          mail_exchange = Domain_name.host_exn @@ Domain_name.of_string_exn domain_name
test/client.ml:      Alcotest.(check bool __LOC__ true @@ Dns.Rr_map.Mx_set.equal mx_set @@ Dns.Rr_map.Mx_set.of_list @@
eio/client/dns_client_eio.mli:        Eio_main.run @@ fun env ->
eio/client/dns_client_eio.mli:        Dns_client_eio.run @@ fun stack ->
eio/client/ohost.ml:    msgf @@ fun ?header ?tags fmt -> with_stamp header tags k ppf fmt
eio/client/ohost.ml:  Eio_main.run @@ fun env ->
eio/client/ohost.ml:  Dns_client_eio.run env @@ fun stack ->
eio/client/dns_client_eio.ml:    ; timeout = Eio.Time.Timeout.v stack.mono_clock @@ Mtime.Span.of_uint64_ns timeout
eio/client/dns_client_eio.ml:      let ns = to_ip_port @@ nameserver_ips t in
eio/client/dns_client_eio.ml:              (to_ip_port @@ nameserver_ips t)
eio/client/dns_client_eio.ml:          let packet, recv_data = Cstruct.split recv_data @@ packet_len + 2 in
eio/client/dns_client_eio.ml:          if not @@ IM.is_empty ctx.requests then handle_data recv_data else ()
eio/client/dns_client_eio.ml:      if not @@ IM.is_empty ctx.requests then
```
</details>

hannesm:
thanks. about `@@` - you seem to use some branch that is not main for grepping. the eio-dns-client PR is still work in progress as far as I understand.

PR #339, opened 2023-05-31
use the right key for the secondary and their zone transfers

hannesm:
//cc @reynir @dinosaure


reynir (approved):
This solves the issue we observed. The solution makes sense to me, but it was a bit hard to read due to the many shadowing and re-use of variable names. I can look at that in a follow up PR - let's merge this :)

PR #342, opened 2023-07-23
be consistent in the use of Logs vs Log module

palainp:
Dear developpers, the goal for this PR is to print a more specific message in the logs when the library (e.g. so far it prints `DBG [application]` sometimes and `DBG [more-specific]` othertimes, now it always provide the source module in the logs).

I just kept `app/*` with `Logs.*` as it will be displayed as `[application]` and it seems fine.

### DIFF
--- a/resolver/dns_resolver.ml
+++ b/resolver/dns_resolver.ml
@@ -8,6 +8,9 @@ type key = [ `raw ] Domain_name.t * Packet.Question.qtype
 
 let pp_key = Dns_resolver_cache.pp_question
 
+let src = Logs.Src.create "dns-resolver" ~doc:"DNS resolver"
+module Log = (val Logs.src_log src : Logs.LOG)
+
 (* The cache (a Map!?) for answers: once a specific type/name comes in, we know
    which questions can progress now *)
 module QM = Map.Make(struct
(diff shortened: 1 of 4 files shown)


resolver/dns_resolver.ml
```diff
@@ -8,6 +8,9 @@ type key = [ `raw ] Domain_name.t * Packet.Question.qtype
 
 let pp_key = Dns_resolver_cache.pp_question
 
+let src = Logs.Src.create "dns-resolver" ~doc:"DNS resolver"
```
hannesm, line 11:
```suggestion
let src = Logs.Src.create "dns_resolver" ~doc:"DNS resolver"
```

let's stay consistent across this repository.

resolver/dns_resolver_cache.ml
```diff
@@ -4,6 +4,9 @@ open Dns
 
 module N = Domain_name.Set
 
+let src = Logs.Src.create "dns-resolver-cache" ~doc:"DNS resolver cache"
```
hannesm, line 7:
```suggestion
let src = Logs.Src.create "dns_resolver_cache" ~doc:"DNS resolver cache"
```

resolver/dns_resolver_utils.ml
```diff
@@ -3,6 +3,9 @@
 open Dns
 open Dns_resolver_cache
 
+let src = Logs.Src.create "dns-resolver-util" ~doc:"DNS resolver util"
```
hannesm, line 6:
```suggestion
let src = Logs.Src.create "dns_resolver_util" ~doc:"DNS resolver util"
```

hannesm:
thanks a lot

resolver/dns_resolver.ml
palainp: Yes of course, sorry for that.

PR #347, opened 2023-11-28
secondary server, mirage layer

hannesm:
If there's a (Dns_server.Secondary.)timer-triggered connection request, do not infinitely try to connect (via request -> fail -> close -> request) in a tight loop, but cancel on first sight (the timer is responsible for re-connecting)

I'm in the process of testing this in production.

### DIFF
--- a/mirage/server/dns_server_mirage.ml
+++ b/mirage/server/dns_server_mirage.ml
@@ -178,7 +178,7 @@ module Make (P : Mirage_clock.PCLOCK) (M : Mirage_clock.MCLOCK) (TIME : Mirage_t
       end
     in
 
-    let rec close ip =
+    let rec close ?(timer = false) ip =
       (match Ipaddr.Map.find_opt ip !tcp_out with
        | None -> Lwt.return_unit
        | Some f -> T.close f) >>= fun () ->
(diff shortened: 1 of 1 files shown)


hannesm:
I think the entire dns_server_mirage and dns_mirage logic may need an overhaul with a view on when which failures may arise, and how to properly handle them (to avoid such tight loops). maybe `timer` is the only thing, though...

mirage/server/dns_server_mirage.ml
```diff
@@ -187,12 +187,15 @@ module Make (P : Mirage_clock.PCLOCK) (M : Mirage_clock.MCLOCK) (TIME : Mirage_t
       let elapsed = M.elapsed_ns () in
       let state', out = Dns_server.Secondary.closed !state now elapsed ip in
       state := state' ;
-      request (ip, out)
-    and read_and_handle ip f =
+      if not timer then
+        request ~timer (ip, out)
+      else
+        Lwt.return_unit
```
reynir, line 193: At an earlier iteration this was clearer, but `close` is only called with `~timer:true` through `request ~timer:true` (again through `timer`). We could move this code into `request`:

```OCaml
close _ >>= fun () ->
if timer then Lwt.return_unit else request _
```

I had written this comment once before but it seems it was not submitted(?!)

hannesm: Hmm?

`close` is called from various points in the code, and the `~timer` can be true or false (indicating whether the connection establishment originated from the timer). Now, when `timer` is false, then `close` calls `request` again. (I'm still not sure whether this is wise -- but there are some (i.e. let's-encrypt-secondary) unikernels that rely on the fact that the TCP connection is always established (to receive NOTIFY events)...

reynir: Nevermind, this is not the case anymore since I believe 8e152f66b440de7716d148d59cebb626cfadea85.

reynir:
I think this improves the situation so I will merge. If there are unintended effects we can revisit.

PR #348, opened 2023-12-05
The b root name server IP was changed on 2023-11-27

hannesm:
See the announcement https://www.lacnic.net/6869/2/lacnic/lacnic-assigns-number-resources-to-the-usc_isi-dns-root-server


reynir (approved):
I confirmed that I get the new addresses when querying NS for `.`. Thanks!

PR #349, opened 2024-04-28
add a packet callback

RyanGibb:
Which is useful for custom processing, e.g. DNS load balancing or [DNS tunneling](https://github.com/RyanGibb/eon/tree/main/lib/transport).


server/dns_server.mli
```diff
@@ -102,6 +102,9 @@ val handle_tsig : ?mac:Cstruct.t -> t -> Ptime.t -> Packet.t ->
 (** [handle_tsig ~mac t now packet buffer] verifies the tsig
     signature if present, returning the keyname, tsig, mac, and used key. *)
 
+type packet_callback = Packet.t -> Packet.t option
+(** [callback question] either returns a faux-handled DNS query [Some answer] or [None]. *)
```
hannesm, line 106: should its type be more constrained, i.e. `Packet.request -> Packet.reply option`? I'm struggling a bit what a good API would be -- and what to allow to clients of this API?

From the DNS server perspective, the invariant to satisfy is that each decoded request should be answered with a reply. This also means that stuff like ID in the header etc. should be good (so no client is confused) -- also fragmentation (including edns information), edns reply, ... would be nice to handle only at one point, wouldn't it?

What do you think? What is the information you really need in your use case? And would it simplify your code if you'd just reply with an answer (instead of a whole packet)?

server/dns_server.mli
```diff
@@ -141,13 +144,13 @@ module Primary : sig
...
-  val handle_buf : s -> Ptime.t -> int64 -> proto ->
+  val handle_buf : ?packet_callback:packet_callback -> s -> Ptime.t -> int64 -> proto ->
     Ipaddr.t -> int -> Cstruct.t ->
     s * Cstruct.t list * (Ipaddr.t * Cstruct.t list) list *
     [ `Notify of Soa.t option | `Signed_notify of Soa.t option | `Keep ] option *
     [ `raw ] Domain_name.t option
-  (** [handle_buf s now ts proto src src_port buffer] decodes the [buffer],
-     processes the DNS frame using {!handle_packet}, and encodes the reply.
+  (** [handle_buf ~packet_callback s now ts proto src src_port buffer] decodes the
+     [buffer], processes the DNS frame using {!handle_packet}, and encodes the reply.
```
hannesm, line 153: in the docstring, the packet_callback should be mentioned and its semantics. but this is best done after we agree on the type of the callback.

server/dns_server.mli
RyanGibb: > should its type be more constrained, i.e. Packet.request -> Packet.reply option? I'm struggling a bit what a good API would be -- and what to allow to clients of this API?

That sounds like a good idea. What about Packet.Question.t -> Packet.reply? This would allow the client to see that name that's being queried.

hannesm: For me, `Question.t -> reply option` would be great. If that is all you need, perfect.

RyanGibb: Great! I'll change the commit to reflect this.

hannesm:
Sorry for the long pause, I pushed a commit which adds a bit more documentation. If this reads good for you, I'm happy to merge.

RyanGibb:
Thanks, Hannes, this is great!

PR #350, opened 2024-04-28
expose Dns_server.update_data

RyanGibb:
For out-of-band updates with the same semantics as DNS UPDATE. I've experimenting with a [capnproto interface](https://github.com/RyanGibb/eon/blob/839747770685f3553899df40ad799d6bb139803e/lib/cap/schema.capnp#L62) that [uses](https://github.com/RyanGibb/eon/blob/839747770685f3553899df40ad799d6bb139803e/lib/cap/domain.ml#L339) this function.

### DIFF
--- a/server/dns_server.mli
+++ b/server/dns_server.mli
@@ -76,6 +76,13 @@ val handle_question : t -> Packet.Question.t ->
 (** [handle_question t question] handles the DNS query [question] by looking
     it up in the trie of [t]. *)
 
+val update_data : Dns_trie.t -> 'a Domain_name.t ->
+  Dns.Packet.Update.prereq list Domain_name.Map.t
+  * Dns.Packet.Update.update list Domain_name.Map.t ->
+  ( Dns_trie.t * (Domain_name.Set.elt * Dns.Soa.t) list,
+   Dns.Rcode.t )
+  result
+
 val handle_update : t -> proto -> [ `raw ] Domain_name.t option ->
   Packet.Question.t -> Packet.Update.t ->
   (Dns_trie.t * ([`raw] Domain_name.t * Soa.t) list, Rcode.t) result


server/dns_server.mli
```diff
@@ -76,6 +76,13 @@ val handle_question : t -> Packet.Question.t ->
 (** [handle_question t question] handles the DNS query [question] by looking
     it up in the trie of [t]. *)
 
+val update_data : Dns_trie.t -> 'a Domain_name.t ->
+  Dns.Packet.Update.prereq list Domain_name.Map.t
+  * Dns.Packet.Update.update list Domain_name.Map.t ->
+  ( Dns_trie.t * (Domain_name.Set.elt * Dns.Soa.t) list,
+   Dns.Rcode.t )
+  result
```
hannesm, line 84: there's need for a docstring describing the semantics -- and what will be done, what won't be (e.g. authentication).

also, the return value should be described in detail -- what the second projection is supposed to contain, etc.

I also wonder what the operation for a DNS server should be -- so once an update_data has been performed, there may be need to notify secondary servers -- how's this supposed to happen? and with your cap'n'proto work, how do you notify the other servers when a DNS update is performed?

hannesm:
I'm really uncertain about the semantics this should have, in terms of notifying other servers (since the notification mechanism is not exposed via API as far as I know, a potential client could use this proposed `update_data` and get out-of-sync name servers).

Maybe one path is to refactor the handle_packet function, but I just looked into it, and there's as well the incremental zone transfer support (update_trie_cache), and the notification mechanism.

On the other hand, we can move along and just merge it -- there's already the `handle_update` exposed, and `with_data`. Maybe there just needs to be more clear documentation at these functions.

RyanGibb:
> I'm really uncertain about the semantics this should have, in terms of notifying other servers
> On the other hand, we can move along and just merge it -- there's already the `handle_update` exposed

Yep, I think `handle_update` has the same issue.

> and with your cap'n'proto work, how do you notify the other servers when a DNS update is performed?

I haven't added this yet, but one option would be to propagate all updates over the RPC interface Though this might be a bit wasteful if the prerequisite conditions aren't met. If I'm reading the code correctly `handle_update` should return an `Error` if they aren't met, so we could forward the updates conditional on an `Ok` result.

PR #351, opened 2024-04-29
remove cstruct from dns, dns-client, dns-client-lwt

hannesm:
still WIP since there are some cstructs around

### DIFF
--- a/app/dns_cli.ml
+++ b/app/dns_cli.ml
@@ -15,12 +15,12 @@ let connect_tcp ip port =
 (* TODO EINTR, SIGPIPE *)
 let send_tcp sock buf =
   let size = String.length buf in
-  let size_cs =
+  let size_buf =
     let b = Bytes.create 2 in
     Bytes.set_int16_be b 0 size ;
     b
   in
-  let data = Bytes.cat size_cs (Bytes.of_string buf) in
+  let data = Bytes.cat size_buf (Bytes.of_string buf) in
   let whole = size + 2 in
   let rec out off =
     if off = whole then ()
--- a/app/dns_cli.ml
+++ b/app/dns_cli.ml
@@ -14,13 +14,13 @@ let connect_tcp ip port =
 
 (* TODO EINTR, SIGPIPE *)
 let send_tcp sock buf =
-  let size = Cstruct.length buf in
+  let size = String.length buf in
   let size_cs =
-    let b = Cstruct.create 2 in
-    Cstruct.BE.set_uint16 b 0 size ;
+    let b = Bytes.create 2 in
+    Bytes.set_int16_be b 0 size ;
     b
   in
-  let data = Cstruct.(to_bytes (append size_cs buf)) in
+  let data = Bytes.cat size_cs (Bytes.of_string buf) in
   let whole = size + 2 in
   let rec out off =
     if off = whole then ()
@@ -39,10 +39,10 @@ let recv_tcp sock =
   in
   let buf = Bytes.create 2 in
   read_exactly buf 2 0 ;
-  let len = Cstruct.BE.get_uint16 (Cstruct.of_bytes buf) 0 in
+  let len = Bytes.get_int16_be buf 0 in
   let buf' = Bytes.create len in
   read_exactly buf' len 0 ;
-  Cstruct.of_bytes buf'
+  Bytes.unsafe_to_string buf'
 
 open Cmdliner
 
--- a/mirage/client/dns_client_mirage.ml
+++ b/mirage/client/dns_client_mirage.ml
@@ -33,7 +33,7 @@ module type S = sig
 end
 
 module Make
-  (R : Mirage_random.S)
+  (R : Mirage_crypto_rng_mirage.S)
   (T : Mirage_time.S)
   (M : Mirage_clock.MCLOCK)
   (P : Mirage_clock.PCLOCK)
@@ -171,7 +171,7 @@ The format of a nameserver is:
         if retries = 0 then
           Error (`Msg "couldn't find a free UDP port")
         else
-          let port = 1024 + ((Cstruct.BE.get_uint16 (R.generate 2) 0) mod (65536 - 1024)) in
+          let port = 1024 + ((String.get_uint16_be (R.generate 2) 0) mod (65536 - 1024)) in
           if IS.mem port t.udp_ports then
             go (retries - 1)
           else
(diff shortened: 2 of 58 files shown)


dinosaure:
I completed this PR with a fix for all other packages including unikernels. For the latter, we still need to convert everything to string dynamically (due to constraints on interfaces) but the code does not depends on `cstruct` anymore :tada:. The tests works for me too. The PR needs `pin-depends`, not sure if I should add them or not (or waiting the release train).

src/dns.ml
```diff
@@ -2873,17 +2874,17 @@ module Rr_map = struct
       else String.sub s 0 pos ^ " " ^ ws_after_56 (String.sub s pos (l - pos))
     in
     let hex cs =
-      let buf = Bytes.create (Cstruct.length cs * 2) in
-      for i = 0 to pred (Cstruct.length cs) do
-        let byte = Cstruct.get_uint8 cs i in
+      let buf = Bytes.make (String.length cs * 2) '\000' in
```
palainp, line 2877: Below exactly 2 bytes are written for each loop iteration, so `Bytes.create` should be ok too.

src/dns.ml
```diff
@@ -4726,29 +4727,29 @@ module Packet = struct
...
+    if String.length buf < 12 then
       None
     else
-      let query = Cstruct.get_uint8 buf 2 lsr 7 = 0 in
+      let query = String.get_uint8 buf 2 lsr 7 = 0 in
       if not query then (* never reply to an answer! *)
         None
       else
-        let hdr = Cstruct.create 12 in
+        let hdr = Bytes.make 12 '\000' in
```
palainp, line 4737: Here it seems that only a subpart is initialized (bytes 0-4 and 10-11 in the else alternative), so probably it's better to keep `Bytes.make` :)

src/dns.ml
```diff
@@ -4667,23 +4668,23 @@ module Packet = struct
...
-          Cstruct.length buf, true
+          Bytes.set_uint8 buf 2 (0x02 lor (Bytes.get_uint8 buf 2)) ;
+          Bytes.length buf, true
       in
-      Cstruct.sub buf 0 off, trunc
+      String.sub (Bytes.unsafe_to_string buf) 0 off, trunc
     in
     let rec doit s =
-      let cs = Cstruct.create s in
+      let cs = Bytes.make s '\000' in
```
palainp, line 4687: I've not reviewed carefully this one, nor the following one :(

app/dns_cli.ml
```diff
@@ -14,13 +14,13 @@ let connect_tcp ip port =
 
 (* TODO EINTR, SIGPIPE *)
 let send_tcp sock buf =
-  let size = Cstruct.length buf in
+  let size = String.length buf in
   let size_cs =
-    let b = Cstruct.create 2 in
-    Cstruct.BE.set_uint16 b 0 size ;
+    let b = Bytes.create 2 in
+    Bytes.set_int16_be b 0 size ;
```
palainp, line 20: As it's not anymore a cstruct, maybe rename to size_buf or equivalent?

palainp (commented):
Thank you for that huge work.
I have a few comments, mainly about `Bytes.create` vs `Bytes.make`, these may be premature optimisations. I think at least one of them is still mandatory.
This is just nitpicking, but perhaps it's worth renaming `*_cs` to `*_bytes`?

mirage/client/dns_client_mirage.ml
```diff
@@ -211,7 +211,7 @@ The format of a nameserver is:
       }
 
     let nameservers { proto ; nameservers ; _ } = proto, nameservers
-    let rng = R.generate ?g:None
+    let rng n = Cstruct.to_string (R.generate ?g:None n)
```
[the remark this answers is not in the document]
hannesm: And we can even could the middle man, and use `Mirage_crypto_rng_mirage.S`. The `mirage-random` should just vanish...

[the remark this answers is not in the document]
dinosaure: I currently prefer to do this work in another PR. The implications of such a change, particularly as regards unikernels, have not been fully established. At this stage and in view of the release train, it's better to concentrate on a _releasable_ state rather than add another migration job.

src/dns.ml
dinosaure: As with `mirage-crypto`, I've been very conservative in my approach (i.e. to adopt the same behavior as `Cstruct`) because I've reported test failures due to uninitialized buffers. In my opinion, we should keep this conservative approach (and avoid a few surprises like what happened with [x509](https://github.com/mirage/mirage-crypto/pull/226)) unless we can really establish that everything is initialized correctly.

src/dns.ml
```diff
@@ -2014,85 +2012,85 @@ module Tsig = struct
...
 
   let encode_raw_tsig_base name t =
     let name = name_to_buf (Domain_name.canonical name)
     and aname = name_to_buf (algorithm_to_name t.algorithm)
     in
-    let clttl = Cstruct.create 6 in
-    Cstruct.BE.set_uint16 clttl 0 Class.(to_int ANY_CLASS) ;
-    Cstruct.BE.set_uint32 clttl 2 0l ;
-    let time = Cstruct.create 8 in
+    let clttl = Bytes.make 6 '\000' in
```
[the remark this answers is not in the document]
palainp: this one was forgotten :)

src/dns.ml
palainp: Yes I totally approve, that's why I tried to make sure that all bytes are written when I suggested switching to `Bytes.create`. During the review I left this one, and the following one unreviewed, sorry :(
BTW I'm unsure about `let hdr = Bytes.create 12 in' in` at line 4737, as I haven't find where all bytes are initialized :(

src/dns.ml
palainp: Sorry for the first comment, I just wanted to comment on each `Bytes.make`, and here I think it just added noise. This `Bytes.make` was replaced in https://github.com/mirage/ocaml-dns/pull/351/commits/68a823e7e7652bf4e6067d0ea3e6fea0906553fd among all that similar replacements. But, as I understand it, it shouldn't be since only a few bytes are defined.

mirage/client/dns_client_mirage.ml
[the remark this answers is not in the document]
hannesm: I only read your comment now, but pushed fdaa332f here. we can of course separate that commit into separate PR if you prefer.

PR #352, opened 2024-05-03
Provide a Miou implementation of dns-client

dinosaure:
A free implementation of `dns-client` with Miou (and `happy-eyeballs-miou-unix`).

### DIFF
--- a/dns-client-miou-unix.opam
+++ b/dns-client-miou-unix.opam
@@ -27,7 +27,3 @@ synopsis: "DNS client API for Miou"
 description: """
 A client implementation using uDNS using Miou.
 """
-
-pin-depends: [
-  [ "happy-eyeballs-miou-unix.1.1.0" "git+https://github.com/robur-coop/happy-eyeballs.git#2a2f276b547d9ac34c0ce9ca40a19b67c7db726f" ]
-]
--- a/dns-client-miou-unix.opam
+++ b/dns-client-miou-unix.opam
@@ -28,7 +28,5 @@ A client implementation using uDNS using Miou.
 """
 
 pin-depends: [
-  [ "miou.dev" "git+https://git.robur.coop/robur/miou.git#ed5087b832797616df073bd8ec9baed2ec4e474c" ]
-  [ "happy-eyeballs-miou-unix.0.6.0" "git+https://github.com/dinosaure/happy-eyeballs.git#4ff9ff889500b764adcefaf719852298c8d51a40" ]
-  [ "tls-miou-unix.0.17.4" "git+https://github.com/dinosaure/ocaml-tls.git#6b635d130b83bb45007e374d828f877332ab2b2e" ]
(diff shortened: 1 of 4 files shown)


dns-client-miou-unix.opam
```diff
@@ -0,0 +1,33 @@
...
+  "happy-eyeballs-miou-unix"
+]
+synopsis: "DNS client API for Miou"
+description: """
+A client implementation using uDNS using Miou.
+"""
+
+pin-depends: [
+  [ "happy-eyeballs-miou-unix.1.1.0" "git+https://github.com/robur-coop/happy-eyeballs.git#2a2f276b547d9ac34c0ce9ca40a19b67c7db726f" ]
+]
```
hannesm, line 33:
```suggestion
```

hannesm:
from my perspective, this is ready to go!? I just merged the happy-eyeballs to opam-repository and removed the pin-depend here.

dns-client-miou-unix.opam
```diff
@@ -0,0 +1,29 @@
...
+
+build: [
+  [ "dune" "subst"] {dev}
+  [ "dune" "build" "-p" name "-j" jobs ]
+  [ "dune" "runtest" "-p" name "-j" jobs ] {with-test}
+]
+
+depends: [
+  "dune" {>="2.0.0"}
+  "ocaml" {>= "4.08.0"}
```
hannesm, line 17:
```suggestion
  "ocaml" {>= "5.0.0"}
```

miou/client/dns_client_miou_unix.ml
```diff
@@ -0,0 +1,186 @@
...
+            let rx_len' = rx_len - (expected_len + 2) in
+            Bytes.blit buf (expected_len + 2) buf' 0 rx_len';
+            go buf' rx_len'
+        | Some expected_len when Bytes.length buf >= expected_len + 2 ->
+          let len = (expected_len + 2) - rx_len in
+          Tls_miou_unix.really_read fd buf ~off:rx_len ~len;
+          go buf (rx_len + len)
+        | Some expected_len ->
+          let buf' = Bytes.make (expected_len + 2) '\000' in
+          Bytes.blit buf 0 buf' 0 rx_len;
```
hannesm, line 120: I'm actually unsure about `buf'` -- what is it's purpose? we allocate and blit into it, but then we throw it away!?

hannesm: I suspect the line below should use `buf'`. The `go` function is a bit hard to read and understand imho...

dinosaure: If I read correctly, this is the worst case where the current `buf` is not large enough to contains the actual DNS packet. In this branch, the idea is to allocate a new bytes `buf'`) which can fit the DNS packet and use it (and forget the old `buf`). Actually, we have an error on the `go buf rx_len`, it should be `go buf' rx_len`.

dinosaure: Fixed, I added a comment to explain what is the purpose of this branch.

PR #361, opened 2025-01-09
Dnssec improvements

hannesm:


dns-cli.opam
```diff
@@ -23,6 +23,9 @@ depends: [
   "mirage-crypto" {>= "1.0.0"}
   "mirage-crypto-pk" {>= "1.0.0"}
   "mirage-crypto-rng" {>= "1.0.0"}
+  "mirage-unix"
```
hannesm, line 26:
```suggestion
  "mirage-unix" {>= "5.0.0"}
```

dns-cli.opam
```diff
@@ -23,6 +23,9 @@ depends: [
...
+  "mirage-clock-unix"
```
hannesm, line 27:
```suggestion
  "mirage-clock-unix" {>= "4.0.0"}
```

reynir:
Sorry about the force pushes. I pushed some commits I didn't mean to, and then I reverted to the wrong commit. Should be back to what it was earlier today.

hannesm:
let's merge this, and improve later.

PR #373, opened 2025-05-22
Extend a bit the Dns_mirage_resolver API

dinosaure:
It's needed to be able to implement DNS over HTTPS.


hannesm:
the issue is that this new function should act inside of the `resolver` and do the recursive resolution, rather than stand independent... maybe the `val resolver` should return that function instead of a unit?

mirage/resolver/dns_resolver_mirage.mli
```diff
@@ -1,9 +1,14 @@
...
+  type t = Dns_resolver.t ref
 
-  val resolver : S.t -> ?root:bool -> ?timer:int -> ?udp:bool -> ?tcp:bool -> ?tls:Tls.Config.server -> ?port:int -> ?tls_port:int -> Dns_resolver.t -> unit
-  (** [resolver stack ~root ~timer ~udp ~tcp ~tls ~port ~tls_port resolver]
+  val connect : Dns_resolver.t -> t
+  val handle : dst:Ipaddr.t -> port:int -> string -> t ->
+    (Dns.proto * Ipaddr.t * int * string) list * (Dns.proto * Ipaddr.t * string) list
+
+  val resolver : S.t -> ?root:bool -> ?timer:int -> ?udp:bool -> ?tcp:bool -> ?tls:Tls.Config.server -> ?port:int -> ?tls_port:int -> t -> unit
+  (** [deamon stack ~root ~timer ~udp ~tcp ~tls ~port ~tls_port resolver]
```
hannesm, line 11: not sure why it is now `daemon` here.

mirage/resolver/dns_resolver_mirage.ml
```diff
@@ -23,11 +23,24 @@ module Make (S : Tcpip.Stack.V4V6) = struct
         | x -> x
     end)
 
-  let resolver stack ?(root = false) ?(timer = 500) ?(udp = true) ?(tcp = true) ?tls ?(port = 53) ?(tls_port = 853) t =
+  type nonrec t = Dns_resolver.t ref
+
+  let connect (t : Dns_resolver.t) : t = { contents= t }
```
reynir, line 28: I know what this code does, but I more often see people using the `ref` "constructor"
```suggestion
  let connect (t : Dns_resolver.t) : t = ref t
```

mirage/resolver/dns_resolver_mirage.ml
```diff
@@ -23,11 +23,24 @@ module Make (S : Tcpip.Stack.V4V6) = struct
         | x -> x
     end)
 
-  let resolver stack ?(root = false) ?(timer = 500) ?(udp = true) ?(tcp = true) ?tls ?(port = 53) ?(tls_port = 853) t =
+  type nonrec t = Dns_resolver.t ref
```
reynir, line 26: I guess with OCaml 5 we should consider concurrenc/parallelism /o\

reynir (commented):
Nice work!

mirage/resolver/dns_resolver_mirage.mli
```diff
@@ -1,8 +1,9 @@
 (* (c) 2017, 2018 Hannes Mehnert, all rights reserved *)
 
 module Make (S : Tcpip.Stack.V4V6) : sig
-
-  val resolver : S.t -> ?root:bool -> ?timer:int -> ?udp:bool -> ?tcp:bool -> ?tls:Tls.Config.server -> ?port:int -> ?tls_port:int -> Dns_resolver.t -> unit
+  val resolver
+    :  S.t -> ?root:bool -> ?timer:int -> ?udp:bool -> ?tcp:bool -> ?tls:Tls.Config.server -> ?port:int -> ?tls_port:int
+    -> Dns_resolver.t -> ((Ipaddr.t * int) -> string -> string Lwt.t)
```
reynir, line 6: I think the return type deserves a comment!

reynir: Perhaps the return type should be an opaque `type resolver` and we add a function `val resolve_external : (Ipaddr.t * int) -> string -> string Lwt.t` that just calls the function. Then we can more easily change the implementation.

PR #376, opened 2025-06-12
record some statistics about the resolver

hannesm:


reynir:
The experimental windows build fails with dune-project error. The error message seem to indicate `dune-project` is full of NUL bytes. The s390x builds are still waiting for a worker. Otherwise it is green.

PR #380, opened 2025-06-18
Block list

reynir:
This adds some heuristics to the resolver to detect if the domain is blocked when we get a reply from our internal "primary". When we detect it's blocked we extend the reply with the extended dns error "filtered". We also emit a `` `Blocked `` Metrics event.


resolver/dns_resolver.ml
```diff
@@ -254,6 +254,31 @@ let scrub_it t proto zone edns ts ~signed qtype p =
...
+  | `Answer (answ, _auth) ->
+    (* HACK! We assume blocked domains have a certain shape. *)
+    Domain_name.Map.for_all
+      (fun _domain rr ->
+         Rr_map.for_all
+           (function
+             | Rr_map.B (Rr_map.A, (3600l, ips)) ->
+               Ipaddr.V4.Set.equal ips (Ipaddr.V4.(Set.singleton localhost))
+             | Rr_map.B (Rr_map.Aaaa, (3600l, ips)) ->
+               Ipaddr.V6.Set.equal ips (Ipaddr.V6.(Set.singleton localhost))
```
reynir, line 268: The match on 3600 ttl is probably overly specific. I'm also thinking we should check the SOA the same way as in the NXdomain case...

reynir:
> And in case we do that, we can use e.g. the SOA serial as the extended DNS error code?

I don't quite understand this one. Would we use a magical SOA serial for blocked domains?

reynir:
Yes, a SOA would be sufficient. This PR also works in that case assuming the SOA has the primary name server as "localhost". This is what mirage-hole does so I went with the same approach. We can as well use other approaches such as a magical SOA serial or a magical hostname for the primary name server etc etc. and I'm open to that.

Perhaps @jmid has any opinions on that as the author of mirage-hole :-)

resolver/dns_resolver.ml
```diff
@@ -254,6 +254,37 @@ let scrub_it t proto zone edns ts ~signed qtype p =
...
+    Domain_name.Map.cardinal auth > 0 &&
+    Domain_name.Map.for_all (fun _domain rr ->
+        match Rr_map.find Rr_map.Soa rr with
+        | None -> false
+        | Some soa ->
+          Domain_name.equal soa.nameserver (Domain_name.of_string_exn "localhost") ||
+          Domain_name.equal soa.nameserver (Domain_name.of_string_exn "blocked"))
+      auth
+  in
+
```
hannesm, line 269:
```suggestion
```

reynir:
Yes, it's not clear to me which one to use. FWIW pi-hole uses 15 - Blocked (but not in the first reply - you have to query twice for some reason).

PR #384, opened 2025-06-24
resolver: in the extended error, report the SOA hostmaster as filtering reason

hannesm:
/cc @reynir

### DIFF
--- a/resolver/dns_resolver.ml
+++ b/resolver/dns_resolver.ml
@@ -284,6 +284,36 @@ let likely_blocked reply =
   | `Rcode_error (Rcode.NXDomain, _, Some (_answ, auth)) -> blocked_soa auth
   | _ -> false
 
+let blocking_reason reply =
+  let find_soa_hostmaster rr =
+    let soa = Rr_map.find Rr_map.Soa rr in
+    Option.map (fun soa -> Domain_name.to_string soa.Soa.hostmaster) soa
+  in
+  let find_soa_hostmaster_in_domain_map map =
+    Domain_name.Map.fold (fun _domain rr acc ->
+        match acc, find_soa_hostmaster rr with
+        | None, x -> x
+        | Some x, None -> Some x
+        | Some x, Some y ->
+          if not (String.equal x y) then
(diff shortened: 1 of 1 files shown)


resolver/dns_resolver.ml
```diff
@@ -284,6 +284,36 @@ let likely_blocked reply =
   | `Rcode_error (Rcode.NXDomain, _, Some (_answ, auth)) -> blocked_soa auth
   | _ -> false
 
+let blocking_reason reply =
+  let find_soa_hostmaster rr =
+    let soa = Rr_map.find Rr_map.Soa rr in
+    Option.map (fun soa -> Domain_name.to_string soa.Soa.hostmaster) soa
+  in
```
reynir, line 291: In `likely_blocked` we use a heuristic that the soa has a nameserver of either `localhost` or `blocked`. Maybe we should filter on that here too?

resolver/dns_resolver.ml
```diff
@@ -299,12 +329,13 @@ let handle_primary t now ts proto sender sport packet _request buf =
             (* After guessing that a domain is blocked we add [`Filtered]
                extended error code and emit a [`Blocked] metrics event. *)
             let edns =
+              let reason = blocking_reason reply in
```
reynir, line 332: We could add a few more words. For the dnsvizor PR it would only contain the domain nameified URL of the block list. Then again it's hard to know what people might put in the SOA hostmaster in the future.

reynir (commented):
Thanks! I have some comments, but I think this is fine to merge.

resolver/dns_resolver.ml
hannesm: see [babc797](https://github.com/mirage/ocaml-dns/pull/384/commits/babc7974ab98adade32d9d844785a5b5f306211b)

resolver/dns_resolver.ml
hannesm: see [babc797](https://github.com/mirage/ocaml-dns/pull/384/commits/babc7974ab98adade32d9d844785a5b5f306211b)

PR #385, opened 2025-06-24
resolver: Use Blocked EDE over Filtered

reynir:
Initially, Filtered was chosen because it communicated it was filtered due to the client requesting that. That sort of made sense in my head because I was running my own resolver. However, there's no guarantee of that - it may be a guest on my network etc etc.

Furthermore, pi-hole uses Blocked and the description of Blocked is more neutral.

See also previous discussion here: https://github.com/mirage/ocaml-dns/pull/380#issuecomment-3000649566


hannesm:
thanks

PR #389, opened 2025-07-02
dns-resolver-mirage: implement RFC9539 opportunistic TLS

hannesm:
The connection from recursive resolver to authoritative nameserver is now using TLS if possible. We validate the connection using either the ca-certs-nss if possible, or otherwise we pin the public key fingerprint.


hannesm:
relies on https://github.com/ocaml/opam-repository/pull/28123 for the Ca_certs_nss.trust_anchors API.

PR #394, opened 2025-09-23
Dns_stub: provide an interface, which is fairly similar to Dns_resolver_mirage

hannesm:
This allows us to use the stub resolver in DNSvizor.


hannesm:
@dinosaure yes, well, certainly it is still a construction site, I'm not happy with the dozen of packages this repository entails and would like to cut some.

about dns_resolver_shared, it didn't exist before, and I wasn't aware it is useful externally. we can of course make it public. the Dns_resolver_mirage_shared is intended to be used (by e.g. DNSvizor), and thus is public.

PR #396, opened 2025-10-01
Expose a module type instead of a module shared between dns-mirage-{resolver,stub}

### LINKED ISSUE #395: Expose root DNS servers
Spotted into #394, the distribution of `ocaml-dns` has changed since then, particularly the `dns-resolver` package. It is currently not possible to obtain the information provided by the [`Dns_resolver_root`](https://github.com/mirage/ocaml-dns/blob/main/resolver/dns_resolver_root.mli) module outside of the distribution.

There is also the appearance of the `__private__` sub-package, now due to the fact that `dns_resolver_shared` is not explicitly public (in the terms of dune):
```shell
> ocamlfind list|rg dns
dns                 (version: 10.2.1)
dns-certify         (version: 10.2.1)
dns-certify.mirage  (version: 10.2.1)
dns-client          (version: 10.2.1)
dns-client-lwt      (version: 10.2.1)
dns-client-miou-unix (version: 10.2.1)
dns-client-mirage   (version: 10.2.1)
dns-client.resolvconf (version: 10.2.1)
dns-client.unix     (version: 10.2.1)
dns-mirage          (version: 10.2.1)
dns-resolver        (version: 10.2.1)
dns-resolver.__private__ (version: n/a)
dns-resolver.__private__.dns_resolver_shared (version: 10.2.1)
dns-resolver.mirage (version: 10.2.1)
dns-resolver.shared (version: 10.2.1)
dns-server          (version: 10.2.1)
dns-server.mirage   (version: 10.2.1)
dns-server.zone     (version: 10.2.1)
dns-tsig            (version: 10.2.1)
dns.cache           (version: 10.2.1)
dnssec              (version: 10.2.1)
```

One possible solution would be to make `dns_resolver_shared` public (perhaps under the public name `dns-resolver.shared`) here:
https://github.com/mirage/ocaml-dns/blob/3990bab3064ac2ed3dffbb9aad5831ea19f8571c/resolver/dune#L10-L15

However, I find it difficult to understand the integration (and why it was done this way) of `Dns_resolver_shared` into the `Dns_resolver_mirage_shared` module (specially about `modules_without_implementation`):
https://github.com/mirage/ocaml-dns/blob/3990bab3064ac2ed3dffbb9aad5831ea19f8571c/mirage/resolver/dune#L8-L14

However, the `dns-resolver.shared` package is still exposed (but linked to Mirage), and it requires a `type t` to obtain information that can only be constructed using the functor application of `Dns_resolver_mirage` and its `type t` (which replaces the abstract `type t` of `Dns_resolver_mirage_shared` by substitution):
https://github.com/mirage/ocaml-dns/blob/3990bab3064ac2ed3dffbb9aad5831ea19f8571c/mirage/resolver/dns_resolver_mirage.mli#L15

What I don't understand is the intention to hide `dns_resolver_shared` and these modules and only expose them under certain constraints and only within the framework of mirage.

dinosaure:
And expose into dns-resolver.shared few modules such as `Dns_root`, `Dns_metrics` or `Dns_block`.

This PR proposes another layout of the `ocaml-dns` distribution (to fix #395) where we expose few useful elements for the user with the `dns-resolver.shared` package (roots, metrics and how to block). It provides a new package `dns-resolver.mirage.shared` which exposes a signature (and can be re-used into unikernels such as DNSVizor) which is shared between `dns-{resolver,stub}-mirage`.

By this way, we don't have `__private__` sub-libraries and elements which were exposed before are still available. It breaks obviously the API. I'm not sure about name convention.

### DIFF
--- a/mirage/resolver/dns_resolver_mirage.mli
+++ b/mirage/resolver/dns_resolver_mirage.mli
@@ -12,5 +12,5 @@ module Make (S : Tcpip.Stack.V4V6) : sig
       to 853) using the [resolver] configuration. The [timer] is in milliseconds
       and defaults to 500 milliseconds.*)
 
-  include module type of Dns_resolver_mirage_shared with type t := t
+  include Dns_resolver_mirage_shared.S with type t := t
 end
--- a/mirage/resolver/dns_resolver_mirage_shared.ml
+++ b/mirage/resolver/dns_resolver_mirage_shared.ml
@@ -1,0 +1,10 @@
+(* (c) 2017, 2018 Hannes Mehnert, all rights reserved *)
+
+module type S = sig
+  type t
+  
+  val resolve_external : t -> Ipaddr.t * int -> string -> (int32 * string) Lwt.t
+  val primary_data : t -> Dns_trie.t
+  val update_primary_data : t -> Dns_trie.t -> unit
+  val update_tls : t -> Tls.Config.server -> unit
+end
--- a/mirage/resolver/dns_resolver_mirage_shared.mli
+++ b/mirage/resolver/dns_resolver_mirage_shared.mli
@@ -1,20 +1,22 @@
 (* (c) 2017, 2018 Hannes Mehnert, all rights reserved *)
 
-type t
-
-val resolve_external : t -> Ipaddr.t * int -> string -> (int32 * string) Lwt.t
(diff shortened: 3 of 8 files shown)


hannesm:
Thanks.

PR #399, opened 2025-11-17
fixed https/svcb port size encoding

nickbetteridge:
fixed port value length


nickbetteridge:
I'm not too sure why I'm getting an ocaml-ci failure with this - the tests ran perfectly

PR #406, opened 2025-11-26
resolver, stub: add require_domain argument

hannesm:
if there's a query with only a single label (and of kind A or AAAA), we return a no data reply

### DIFF
--- a/mirage/stub/dns_stub_mirage.ml
+++ b/mirage/stub/dns_stub_mirage.ml
@@ -269,7 +269,7 @@ module Make (S : Tcpip.Stack.V4V6) = struct
         | None ->
           let dn, qtyp = question in
           if Domain_name.count_labels dn = 1 && (qtyp = `K (Rr_map.K A) || qtyp = `K (Rr_map.K Aaaa)) then
-            let data = `Answer (Name_rr_map.empty, Name_rr_map.empty) in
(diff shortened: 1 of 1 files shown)


reynir:
What is the motivation for this?

With this the resolver can return a no data reply for a domain that should be nxdomain. I think that's a bit weird.

reynir:
So ignoring the above comment i think the code changes look  fine.

hannesm:
thanks for your question. the motivation is dnsmasq "domain-needed" configuration option. as you can see (defaulta to false), I'm not sure whether it is really worth to support... I guess from a privacy perspective it may be nice (hosts asking for internal host names) -- but then people should just use a local domain name, which will be directly answered by dnsvizor.

so maybe the alternative is to skip this PR (maybe add the documentation changes), and move on (by ignoring the domain-needed flag)

hannesm:
The issue I have with NXDomain is that it should mean "there's nothing below this.". So, if you reply to "type A domain .com" with a NXDomain, the DNS client would think that there's no type NS domain .com, and also nothing below .com -- i.e. no example.com etc.

That's why I'd, if we like this, use NoData instead. Now, this approach proposed here is a partial NoData since we don't even provide a SOA in the authority section.

So, thanks for your question, I'm at the moment at the point where I'd prefer to not merge this, and just not support the `domain-needed`. I'll pick the documentation changes and push them directly.

PR #407, opened 2025-11-27
Query stream

hannesm:
~~Only last commit is relevant.~~ This adds a stream (well, Lwt_condition.t) for queries and answers. ~~This is currently only implemented in the stub (where we have the Lwt monad). The dns-resolver needs some functionality so the dns-resolver-mirage can adhere to the same interface.~~ The implementation covers both stub and resolver.

~~On top of #406;~~ once this PR here is done and adapted to the dns-resolver, it can be used in dnsvizor for the query page.

### DIFF
--- a/mirage/resolver/dns_resolver_mirage.ml
+++ b/mirage/resolver/dns_resolver_mirage.ml
@@ -44,6 +44,7 @@ module Make (S : Tcpip.Stack.V4V6) = struct
     let tls_auth = ref Ipaddr.Map.empty in
     let stream, push = Lwt_stream.create () in
     let opportunistic = List.mem `Opportunistic_tls_authoritative (Dns_resolver.features t) in
+    let queries = Lwt_condition.create () in
 
     let send_tls flow data =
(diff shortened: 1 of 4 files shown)


mirage/resolver/dns_resolver_mirage_shared.mli
```diff
@@ -19,4 +19,7 @@ module type S = sig
   (** [update_tls t tls_config] updates the tls configuration to [tls_config].
       If the resolver wasn't already listening for TLS connections it will
       start listening. *)
+
+  val queries : t -> (Ptime.t * Dns.Packet.Question.t * Ipaddr.t * Dns.Rcode.t * int64 * string) Lwt_condition.t
```
reynir: Given that this is exported I think we should use a record so we have meaningful names to each field. For example it's not clear what the `int64` and the `string` is.

hannesm:
Thanks, I made this independent of #406. Yes, there will be merge conflicts - but they will be easy to navigate :)

PR #421, opened 2026-07-20
Rotate dns results

PizieDust:
This PR implements shuffling for IP addresses but doesn't randomize the entire list.

cc @reynir @hannesm @dinosaure

### DIFF
--- a/src/dns.ml
+++ b/src/dns.ml
@@ -3474,6 +3474,21 @@ module Rr_map = struct
       Bytes.set_uint16_be buf (off' + 4) rdata_len ;
       names, rdata_end
     in
+    let rotate_list lst =                                                   
+      match lst with                                                        
+      | [] | [_] -> lst                                                   
+      | _ ->               
+        let k = Random.int (List.length lst) in                                            
+        let rec split k acc = function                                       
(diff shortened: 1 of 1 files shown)


hannesm:
"but doesn't randomize the entire list" <- would you mind to explain what "the entire list" is?

src/dns.ml
```diff
@@ -3474,6 +3474,21 @@ module Rr_map = struct
...
+        let rec split k acc = function                                       
+          | [] -> List.rev acc                                               
+          | x :: xs as l ->                                                  
+            if k = 0 then                                                    
+              l @ (List.rev acc)                                             
+            else                                                             
+              split (k - 1) (x :: acc) xs                                    
+        in                                                                   
+        split k [] lst   
+      in
```
hannesm, line 3491: I don't quite understand what this "rotate_list" implements. Is this based on some common literature about how to shuffle a list?

PizieDust: No it doesn't really shuffle the list. Just changes where the list starts. I think for DNS round robin it's better to rotate the list, since clients always connect to the first IP in the list. By rotating the list we can be sure all IP's get a fair chance of being picked.

PR #422, opened 2026-07-30
Update README pointing to various git repositories

hannesm:


dinosaure:
I took the liberty to add another link to `annuaire`

